PR #12, opened 2020-06-17
RFC 1002: Optionality qualifier in properties and links

ambv:


text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+  empty sets in fields used in the query.
+
+Downsides
+~~~~~~~~~
+
+* The main disadvantage at time of writing is that the current behavior as of
+  EdgeDB v1.0a3 is the opposite: fields without an explicit qualifier are
+  assumed optional. This poses a migration challenge.
+
+* Required fields have to be always specified at INSERT time and often
```
elprans, line 64: Currently, I can omit an explicit value in an `INSERT` if a required field has a default, i.e `SET DEFAULT` is implied.  What is the motivation behind requiring an explicit `SET DEFAULT`?

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+
+A default value has far reaching consequences which we cover in this
+document.
+
+
+Implementation
+==============
+
+Since the discussion always applies to properties and links, we're referring
+to them simply as *fields* below.
```
1st1, line 34: Let's refer to the as `pointers`. That's the terminology we use for them in schema `schema::Pointer` and in our code.

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Downsides
+~~~~~~~~~
+
+* The main disadvantage at time of writing is that the current behavior as of
+  EdgeDB v1.0a3 is the opposite: fields without an explicit qualifier are
+  assumed optional. This poses a migration challenge.
+
+* Required fields have to be always specified at INSERT time and often
+  specified at UPDATE time, making queries more verbose to write and data
```
1st1, line 65: I don't think mentioning `UPDATE`s is relevant here. Required fields don't necessarily need to be updated when you're updating just for the sake of them being updated.

text/1002-required-by-default.rst
1st1: Yeah, if there's an explicit default for a link/property specified in the schema there should be no need to duplicate it in INSERT (or it defeats the purpose of having defaults in the first place).

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+Since the discussion always applies to properties and links, we're referring
+to them simply as *fields* below.
+
+Assume REQUIRED qualifiers when no explicit qualifier is provided
+-----------------------------------------------------------------
+
+The main advantage of assuming that all fields are required unless otherwise
+specified is its resilience to migration into optionality later. Client code
+that was forced to provide values before will continue working and there are
+no incompatible empty sets in the database prior to the migration.
```
1st1, line 42: That's not quite true. Let's say REQUIRED is the default and you had a schema:

```
type User {
  property name -> str;
  property email -> str;
}
```

and you had a query like this:

```
SELECT (User.name, User.email)
```

that was used to return ALL emails. Note that we can come up with many other queries; this one while impractical still demonstrates the problem.

**Then** you migrate the schema to 

```
type User {
  optional property name -> str;
  property email -> str;
}
```

Now your `SELECT (User.name, User.email)` query works differently. It would omit user emails for users that don't have a name.

So on the surface it is easier to migrate from "required" to "optional", but in practice some existing queries (especially analytical) would in fact break and start returning incorrect results.

On the other hand, migrating from "optional" to "required" would require you to specify a "default".  But then all of the existing queries in your application would continue to work correctly.

Therefore I'm really not sure that the flipping current behavior would be an improvement, quite the opposite.

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+==============
+
+Assume OPTIONAL qualifiers when no explicit qualifier is provided
+-----------------------------------------------------------------
+
+This is the current behavior.
+
+An optional field is a form of incomplete data that allows the application
+to flexibly decide how to proceed. If the data is there, fine. If it isn't,
+the application can include user-level code to deal with that.
```
1st1, line 93: nit: Optionality doesn't always translate into incompleteness. E.g. "User" not having any "friends" (if the link is optional) means simply that they don't have any friends in the system. The data is complete and accurate.

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Downsides
+~~~~~~~~~
+
+Developers might be failing to fill fields at INSERT time due
+to forgetfulness at query creation time, even if the data could be
+provided. They might be harder to fill at a later stage.
+
+Optionality is a special case that the user code has to deal with. Empty sets
+in queries might leave to unexpected results, i.e. missing objects. An empty
```
1st1, line 108: This is a double-edged sword: https://github.com/edgedb/rfcs/pull/12/files#r441843313

text/1002-required-by-default.rst
ambv: > Therefore I'm really not sure that the flipping current behavior would be an improvement, quite the opposite.

Making existing correct queries silently incorrect is indeed a big problem. That, along with the migration, might be enough to reject this entire proposal.

My intuition is that most fields will stay at whatever the *default* is. With the default being `OPTIONAL` it's more likely that people will write queries that unintentionally omit objects with missing pointers. And that's without any migrations, just bad queries from the get go.

You have the most experience with this. How do you write queries and computables? Are you doing defensive programming against empty sets? Do you come back to the schema to ensure pointers are required?

text/1002-required-by-default.rst
ambv: True, multi links are multi properties are more intuitive in this sense. If there can be many then 0 is much more likely a valid option.

text/1002-required-by-default.rst
elprans: I agree with Yury. The semantics of empty set make `required` -> `optional` change a potentially much sneakier and nastier thing.  When you make a pointer `required`, you will get explicit errors from the database asking you to provide missing data at the time of the migration, and either set a default on a pointer, or change your `INSERT`s (which would start to fail explicitly).  So I don't really see a significant *resilience* advantage to migration into optionality as opposed to vice-versa.

Another problem with making `required` the default lies with link properties, which are always optional by design.

1st1: > My intuition is that most fields will stay at whatever the default is. With the default being OPTIONAL it's more likely that people will write queries that unintentionally omit objects with missing pointers. And that's without any migrations, just bad queries from the get go.

I assume that people will mostly write queries that selects simple shapes of data (like GraphQL), where dealing with empty sets isn't often necessary or painful.

Analytical queries will be presumably written in a way to account for empty sets, and that's why the current default to "optional" makes sense for them (flipping to "required" is easier).

> You have the most experience with this. 

That would be @elprans and @vpetrovykh :)

> How do you write queries and computables? 

There is a coalesce operator `??` and `?=` to deal with empty sets. The rules of using them aren't extremely intuitive though.

> Are you doing defensive programming against empty sets?

I don't think we need to advocate for "defensive programming", but we do need to have an easy guide on how to best handle empty sets in different scenarios. I don't think we have such a high-level document yet.

> Do you come back to the schema to ensure pointers are required?

Ideally yes, you need to know your schema when writing queries against it.

text/1002-required-by-default.rst
```diff
@@ -0,0 +1,178 @@
...
+specified is its resilience to migration into optionality later. Client code
+that was forced to provide values before will continue working and there are
+no incompatible empty sets in the database prior to the migration.
+
+An extreme form of optionality everywhere is equivalent to the database
+not having a true schema. Unless it's necessary to make a field optional,
+forcing fields to be required ensures that whenever data *can* be specified,
+it won't be forgotten at INSERT or UPDATE time.
+
+Required fields simplify type information. There's two particular problems
```
elprans: This argument would hold better if we disallowed `OPTIONAL` altogether and went full-on algebraic types. I don't think that's realistic, so in real schemas a substantial number of pointers would still be defined as `OPTIONAL` and so these problems would still apply.

text/1002-required-by-default.rst
1st1: > might be enough to reject this entire proposal.

Let's not reject it, but update this RFC to be "Informational" and explain why we have "optional" as default.

elprans: > My intuition is that most fields will stay at whatever the default is. With the default being OPTIONAL it's more likely that people will write queries that unintentionally omit objects with missing pointers. And that's without any migrations, just bad queries from the get go.

This isn't different from SQL or any other database that allows (and defaults to) NULLs, which is to say most of them. With that in mind, I think that most people, who have some familiarity with databases, would be aware of the problem, and the best we can do is improve the ergonomics around handling (and guarding against) missing data.  The coalescing operators are one thing, but there's also the planned addition of the `<REQUIRED _>` cast, which can be used to guard against empty sets in queries where they're really not desired.

tailhook: I disagree that "tuple case" is the main counter-argument against required by default. I think that case should be dealt with regardless of this default.

The issue with a `link` is more important. But I think it's fine if `link` will be required by default and `multi` link will be optional by default.

tailhook:
Another argument *for the required by default* is that we have required by default parameters `<str>$var`. So symmetry with properties and links is very nice. Otherwise, you need more mental space to remember which one is optional and which one is required and will mess around multiple time before remembering (it's very common to mess up with things that have opposite defaults).

text/1002-required-by-default.rst
ambv: > But I think it's fine if link will be required by default and multi link will be optional by default.

I find that would be rather unintuitive.

ambv:
Alright, @1st1, I rebranded and reworded the RFC to explain the current state of things.

text/1002-required-by-default.rst
elprans: I also think that the dissonance between the requiredness of `single` and `multi` is a larger issue than inconsistency with query args and is an argument *against* the original proposal.

text/1002-optionality-qualifier.rst
```diff
@@ -0,0 +1,199 @@
...
+the application can include user-level code to deal with that.
+
+Moreover, multi properties and multi links make the case of an empty set
+intuitively acceptable: since the cardinality allows for any number of
+elements, a state with an empty set is not surprising.
+
+Plenty of examples of existing databases and RPC systems assume optionality
+by default. Protocol buffers removed support for required fields altogether
+in version 3. JSON-based NoSQL databases like MongoDB assume optionality.
+HTML form fields are optional by default.
```
1st1, line 52: I'd add that relational databases assume optionality by default too.

text/1002-optionality-qualifier.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Moreover, multi properties and multi links make the case of an empty set
+intuitively acceptable: since the cardinality allows for any number of
+elements, a state with an empty set is not surprising.
+
+Plenty of examples of existing databases and RPC systems assume optionality
+by default. Relational databases assume optionality unless a NOT NULL
+constraint is passed. Protocol buffers removed support for required fields
+altogether in version 3. JSON-based NoSQL databases like MongoDB assume
+optionality. HTML form fields are optional by default.
```
1st1, line 53: I'd also drop the nod to HTML forms -- pretty much no one uses them as is in modern apps. Don't get me wrong, the argument is still valid, it just feels a bit out of context.

text/1002-optionality-qualifier.rst
```diff
@@ -0,0 +1,200 @@
...
+make required-to-optional pointer migrations in the future.  Those hide
+a specific gotcha: pre-existing queries that were previously guaranteed
+never to encounter empty sets now will.  Most of the time this will be
+harmless but some queries that are analytical in nature will now fail to
+return objects with empty pointers, which might lead to bugs in user code.
+
+Thrift documentation explains that "required is forever". What they mean
+by this is that required fields have to be provided by the caller, say a
+mobile device. Changing a required field into an optional field requires
+for the database to be updated first before any application code can use
```
1st1, line 140: I'd change the "database" to something related to RPC -- that would make it easier to parse this paragraph. We first discuss the problematic migration for RPCs and then try to draw a parallel with databases.

```suggestion
for the RPC backend to be updated first before any application code can use
```

text/1002-optionality-qualifier.rst
```diff
@@ -0,0 +1,200 @@
...
+outside of the declared type.
+
+Required pointers by default make it more likely that users would have to
+make required-to-optional pointer migrations in the future.  Those hide
+a specific gotcha: pre-existing queries that were previously guaranteed
+never to encounter empty sets now will.  Most of the time this will be
+harmless but some queries that are analytical in nature will now fail to
+return objects with empty pointers, which might lead to bugs in user code.
+
+Thrift documentation explains that "required is forever". What they mean
```
1st1, line 137: Maybe move this entire paragraph into the "A database is not a peer-to-peer RPC platform" section?

PR #13, opened 2020-06-17
RFC1000: Add description of migration history and migration file format

elprans:


text/1000-migrations.rst
```diff
@@ -40,6 +40,10 @@ the necessary transformations using DDL and, possibly, regular queries.
 In the database, the migrations are recorded as ``sys::Migration`` objects::
 
     type sys::Migration {
+        # Migration hash derived from contents and parent hashes.
+        required property hash -> str {
+            constraint exclusive;
+        };
```
1st1, line 46: I don't see the `name` property and the discussion on how migration names would be derived from hashes. Also how one uses the `DESCRIBE` or `DESCRIBE MIGRATION` commands with migration hashes.

text/1000-migrations.rst
```diff
@@ -76,6 +80,41 @@ whether there are untracked changes in the local schema that require a
...
+Migration History
+=================
+
+The migration history is defined as a linear sequence of migrations (although
+this might change in the future to support merging migration histories).
+Each migration has a unique identifier that is a hash derived from the token
+stream of the migration script and the hash of the migration parent.  The
+hashes are computed by the ``edgedb migrate`` command starting from the
+latest committed migration (if any).  It is an error to amend an
+already-committed migration.
```
1st1, line 92: Also need to add that when a database is created, it has a single migration with a null-hash, e.g. a migration with a sole `CREATE DATABASE db;` command.

elprans: > migration with a sole `CREATE DATABASE db;` command.

Migrations cannot contain `DATABASE` commands (also `ROLE`, `MIGRATION`, `TRANSACTION`, etc).

Do we really need an explicit nil migration?  An empty parent would serve the same purpose.

text/1000-migrations.rst
elprans: Do we still want to have separate "hash" and "name" fields?  Arbitrary migration names?

1st1: I'd be fine with `DESCRIBE MIGRATION "hash"` variant and not having `name` field for migrations at all. But I think you wanted them to have names, no?

text/1000-migrations.rst
1st1: > Migrations cannot contain DATABASE commands (also ROLE, MIGRATION, TRANSACTION, etc).

Alright, let's not break the rules here. I guess an implicit null-migration is OK.

text/1000-migrations.rst
elprans: OK, I think we should just use the `name` property, which is generated using the hash.

1st1: Hm, I'd keep the hash as a separate property. Maybe we define `name` as a required property with a computable default off the hash?

text/1000-migrations.rst
```diff
@@ -76,6 +80,71 @@ whether there are untracked changes in the local schema that require a
...
+   Scenarios:
+   a) A user has two in-progress migration files: 0004.edgeql and 0005.edgeql.
+      If the user modifies 0004.edgeql and runs edgedb migration commit, we
+      compute the name for 0004 and commit it; we then update the parent name
+      in 0005 and commit it too.
+   b) A user has two in-progress migration files: 0004.edgeql and 0005.edgeql.
+      If the user modifies 0005.edgeql and runs edgedb migration commit,
+      we first commit the 0004, validate that 0005 has the correct parent
+      name, and if needed we fix it. We also compute it's new name, and then
+      commit 00005.
```
1st1, line 116:
```suggestion
      commit 0005.
```

text/1000-migrations.rst
```diff
@@ -76,6 +80,71 @@ whether there are untracked changes in the local schema that require a
...
+stream of the migration script and the migration parent name.  To
+make migration names valid identifiers the hash is prefixed with letter
+``'m'``.
+
+The names of the migrations are computed by the ``edgedb migrate`` command
+starting from the latest committed migration (if any), using the following
+algorithm:
+
+1. Get the latest migration id from the database.
+2. Scans all files in the migrations dir and finds a file that references
```
1st1, line 98:
```suggestion
2. Scan all files in the migrations dir and finds a file that references
```

text/1000-migrations.rst
```diff
@@ -76,6 +80,71 @@ whether there are untracked changes in the local schema that require a
...
+this RFC is defined as follows.
+
+Each migration is stored in a separate file.  Files are named using a
+simple monotonic decimal counter ('00001.edgeql', '00002.edgeql', etc).
+The contents of the file is the body of the migration and must be valid
+EdgeQL.  Migration name, the name of the parent migration, the
+migration description message and other potential metadata are recorded
+in a comment block at the top of the file:
+
+    # migration: <migration-name>
```
1st1, line 138: Would be cool if we could stick 'autogenerated' somewhere to highlight that `migration` and `parent` fields aren't supposed to be edited.

What about updates to the migration message? Would changes to it be detected and synced to the DB?

text/1000-migrations.rst
elprans: Hm, I think this arrangement makes things more confusing than not.  We refer to parent migrations by hash, but use "name" everywhere else?  I think we should stick with a single identifying property and disallow anything but a hash-derived value in it.

text/1000-migrations.rst
elprans: I think any modification to the already-committed migration should at least be an action that is confirmed by the user, but other than that I see no harm in being able to sync non-critical migration data to the db.

text/1000-migrations.rst
1st1: The name essentially becomes hash, hence don't expect that we ever be able to have custom names.

text/1000-migrations.rst
1st1: >  but other than that I see no harm in being able to sync non-critical migration data to the db.

I wasn't suggesting it's harmful, I was suggesting that we don't know how to easily detect such changes -- the hash does not include the message.

text/1000-migrations.rst
1st1: I also don't like that names essentially invites people to circumvent the hashing of migrations and the integrity checks that hashing enables. I really don't like this whole "hash" is a "name" business.

text/1000-migrations.rst
```diff
@@ -76,6 +80,71 @@ whether there are untracked changes in the local schema that require a
...
+   then there's only one in-progress migration file, so it is committed.
+5. If there's a file with that parent and other files after it it means
+   that we have a few in-progress (not yet committed) migration files.
+   The parent names in them are fluid -- generated and present in the file,
+   but aren't yet in the database.
+   Scenarios:
+   a) A user has two in-progress migration files: 0004.edgeql and 0005.edgeql.
+      If the user modifies 0004.edgeql and runs edgedb migration commit, we
+      compute the name for 0004 and commit it; we then update the parent name
+      in 0005 and commit it too.
```
tailhook, line 111: This shouldn't silently update parent commit and apply. We need a flag to do that ``edgedb migrate --allow-rebasing``. And should fail by default.

The reason is this scenario:
1. I've edited file and did not run ``edgedb migrate`` on my laptop
2. Committed my edited file to git
3. Then CI runs ``edgedb migrate`` on production database, silently modifies the migration files and they are never committed from CI
4. Now revisions in git and production are out of sync.

In this scenario, hopefully you can run migrate locally again and get the same results. But you already don't know which ones are on production and whether there were any extra edits after that.

To fix that we should fail at step 3 and force developers to run `--allow-rebase` on their machine.

text/1000-migrations.rst
tailhook: > I wasn't suggesting it's harmful, I was suggesting that we don't know how to easily detect such changes -- the hash does not include the message, or does it?

If it does, we need another hash to detect that script is the same and it's safe to update metadata (git by the way has both: hash of contents of the commit (tree) and commit hash with metadata, and this allows certain scenarios easier)

elprans: I don't think we should include message and other stuff in the hash, only the parent and the script, so an extra hash doesn't seem to be necessary.

text/1000-migrations.rst
elprans: > I also don't like that names essentially invites people to circumvent the hashing of migrations

I changed my mind re custom names.  They're unnecessary and complicate things, because now you how TWO ways to identify and refer to a migration (a hash and a name).  I think we should settle on one, system-maintained.  Calling it "name" makes things much simpler for the implementation, because all other objects are identified by "name" and lots of code is written with this assumption.

1st1: > I changed my mind re custom names.

Good.

I still don't like the proposed schema design though. It creates an impression that names *can be custom*. Can we do this:

```
type sys::Migration {
        # Migration name derived from the hash of contents and parent name.
        required property hash -> str {
            constraint exclusive;
        };
        required property name  := .hash;
}
```

?

text/1000-migrations.rst
1st1: > And should fail by default.

Should the entire operation fail, i.e. both 0004 and 0005 not committed?

text/1000-migrations.rst
1st1: > so an extra hash doesn't seem to be necessary.

But it is necessary, otherwise we can't detect changes to the migration message on the disk and sync them with the DB.  A metadata hash that @tailhook suggests makes sense to me.

text/1000-migrations.rst
elprans: This arrangement would require hacks on the reflection side, because internally we would still need to define the field as "name", otherwise we would not be able to rely on the common infrastructure for "named" schema objects, as I pointed out above.  This seems like an unnecessary price to pay for me.

1st1: Would it even be possible to set `name` for the user somehow? I.e. `CREATE MIGRATION` and `COMMIT MIGRATION` will compute the hash/name automatically and there's no need for `SET name := ...` syntax, correct?

elprans: Yes, no need, it's read-only from the user's perspective.

text/1000-migrations.rst
elprans: > But it is necessary, otherwise we can't detect changes to the migration message

Why?  Read the files and run an introspection query and compare.  Am I missing something?

text/1000-migrations.rst
1st1: OK then, case closed.

text/1000-migrations.rst
1st1: Yeah, alright, I guess it doesn't really matter if we introspect and compare messages vs introspect and compare hashes.

The command line tooling can have its own cache with hashes of all migration files. That would simplify (and greatly speedup) detection of changed files / metadata.

text/1000-migrations.rst
tailhook: Yes, if user requested to upgrade to the "head" (last revision). I.e. if we know in advance that we can't fulfill the whole operation there is no reason to partially apply.

If we have command `edgedb migrate --to=0004.sql` then operation should succeed (and *may* give a hint that next revision has different parent).

PR #15, opened 2020-07-14
Update of rfc 1001 according to the impl

tailhook:


text/1001-edgedb-server-control.rst
```diff
@@ -339,12 +348,15 @@ Options
...
+``--to-version``
   Specifies the version of EdgeDB to upgrade to.  If not specified,
   the latest available installed version is used.
 
+``--to-nightly``
+  Specifies that the instance should be upgraded to the nightly version.
+
 ``--nightly``
-  Upgrade to a nightly release.
+  Upgrade all instances currently running nightly.
```
elprans, line 359: Does this upgrade nightlies that cross major versions? I.e. does this upgrade alpha.3 nightly to alpha.4 nightly?

text/1001-edgedb-server-control.rst
```diff
@@ -237,11 +237,8 @@ Options
   The name of the EdgeDB instance.  Must be unique.  If not specified,
   the name ``default`` is used.
 
-``--port=<port-number>``
-  Optionally specifies the port number on which the server should listen.
-
-``--server-options -- <options>``
-  Passes ``edgedb-server`` options verbatim.  Must be the last argument.
```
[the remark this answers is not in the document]
tailhook: I don't know how it should work. Should It be persistent? Or should it be just for single start? There are conflicts here and there:
1. If it's not persistent, how should start without foreground even work?
2. Making it persistent is kinda counter intuitive, as `init` is what creates persistent configuration.
3. Or is it only for foreground?
4. Or if they are persistent but when `--foreground` is specified, should it be just for this invocation, or should they be persisted in service file anyway?
4. And for `--port`, it's also unclear where `edgedb -I` would connect to.

So either they (both `--port` and `--server-options`) should be removed from `start` or documented how they work more thoroughfully. I would wait until there is some clear use case for them, so we can figure out right choices.

(`--startup-options` in `init` are fine)

text/1001-edgedb-server-control.rst
tailhook: Yes. We treat `--nightly` as the latest nightly at all tiimes. I.e. we can't install "alpha.3 nightly".

I've just realized that maybe I'm mistreating nightly. Did you mean that `--version 1-alpha.3 --nightly` would work? Currently, I'm treating `--nightly` and `--version` everywhere as mutually exclusive arguments.

text/1001-edgedb-server-control.rst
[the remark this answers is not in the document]
elprans: In my mind, `--server-options` are a non-persistent one-off for `--foreground` and are useful for the same reason `--foreground` is useful: you want to debug the server by passing logging options or adjusting its behavior in some other way.

text/1001-edgedb-server-control.rst
elprans: Current treatment is OK, but isn't completely clear from the formulation in the RFC.

text/1001-edgedb-server-control.rst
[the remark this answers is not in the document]
tailhook: Re-added `--server-options`

text/1001-edgedb-server-control.rst
tailhook: Added a clarification.

PR #18, opened 2020-08-15
RFC1000: Switch to confirm/reject flow, add describe algo

elprans:
This changes how `DESCRIBE CURRENT MIGRATION` works with respect to the
system-proposed DDL.  Instead of generating multiple variants of the
next proposed statement, only one statement is generated, and the client
will have a choice to either accept or reject it by issuing `ALTER
CURRENT MIGRATION REJECT PROPOSED`.  The reason for this change is that
in the general case it is very difficult to come up with a reasonable
set of multiple alternatives of the *next* valid DDL statement, because
different operations have different dependency path, even on relatively
simple deltas.  The new approach sidesteps that problem by not
attempting to present alternatives and instead guides the client with a
series of questions.


text/1000-migrations.rst
```diff
@@ -174,6 +174,20 @@ name is as follows::
 First revision must have a parent version of `initial`.
 
 
+Migration operation classification
+==================================
+
+The migrations system classifies the migrations operations into two categories:
+safe and unsafe, based on whether the operation is automatically reversible
+without losing any prior data.  For example, all ``CREATE`` operations are
+considered safe by definition, but also alterations to schema that doesn't
```
1st1, line 183: The "reversible" word reads ambiguous. I assume it's safe within one migration? Because  if during a migration a user adds commands to copy data to the newly created type it's no longer safe to reverse it.

text/1000-migrations.rst
```diff
@@ -254,8 +268,14 @@ While the migration block is active:
...
-  generated list of statements required to complete the migration.  See
-  the "DESCRIBE MIGRATION" section below for details.
+  generated proposal for the next DDL statement to include to advance
+  the migration.  See the "DESCRIBE MIGRATION" section below for details.
+
+* The ``ALTER CURRENT MIGRATION REJECT PROPOSED`` statement is used to
+  inform the server that the currently proposed statement returned by
+  a prior call to ``DESCRIBE CURRENT MIGRATION AS JSON`` is not acceptable
+  and that the next execution of ``DESCRIBE`` should return an alternative
+  proposal.
```
1st1, line 278: What happens if the space of proposals gets exhausted? Do we allow users to manually enter some DDL?

text/1000-migrations.rst
```diff
@@ -332,47 +349,78 @@ The returned JSON conforms to the following pseudo-schema::
...
+1. Compute schema "A" from the state immediately preceding the
+   ``START MIGRATION`` command plus all statements in the "confirmed" list.
+
+2. Compute schema "B" from the SDL declaration passed to ``START MIGRATION``.
+
+3. Calculate the similarity matrix for each pair of objects of the same class
+   in schema "A" and schema "B" (e.g. compare every object type in schema "A"
+   with schema "B").  Only top-level objects are considered (i.e. object types,
+   abstract link definitions etc). The rows of the matrix represent objects
+   from "A", columns -- objects from "B", and the cells contain
```
1st1, line 386: "The rows of the matrix represent objects from "A", columns -- objects"

not sure I can parse this sentence

text/1000-migrations.rst
```diff
@@ -332,47 +349,78 @@ The returned JSON conforms to the following pseudo-schema::
...
+---------------------------------------
+
+This statement is only valid while inside the ``START MIGRATION`` block.
+When executed it marks the statement found in the "proposed" branch of
+``DESCRIBE CURRENT MIGRATION AS JSON`` as "unacceptable" and forces the
+server to recompute the diff with this command excluded.  This, effectively
+allows the client to implement the (y) / (n) flow for each proposed statement
+in the migration.  For (y) the client would send back the proposed statement
+to be recorded in the "confirmed" list, and for (n) an ``ALTER CURRENT
+MIGRATION REJECT PROPOSED`` is sent instead.
```
1st1, line 422: FWIW this is potentially extendable to multiple-choice.

text/1000-migrations.rst
elprans:
```
+-------------------------+
|      | Foo | Bar | Baz  |
+-------------------------+
|Foo   | 1.0 | 0.2 | 0.1  |
+-------------------------+
|Bar   | 0.2 | 0.9 | 0.3  |
+-------------------------+
```

text/1000-migrations.rst
elprans: Yes

text/1000-migrations.rst
elprans: The word "prior" is key.  The use case here is the ability to rollback a failed deploy, not the ability to rollback a random migration 2 years after the fact.  Otherwise all, but the most trivial changes to the schema would be "safe".

text/1000-migrations.rst
elprans: I added a design discussion note about why multiple-choice in general is hard.

text/1000-migrations.rst
```diff
@@ -254,8 +271,17 @@ While the migration block is active:
...
-  generated list of statements required to complete the migration.  See
-  the "DESCRIBE MIGRATION" section below for details.
+  generated proposal for the next DDL statement to include to advance
+  the migration.  See the "DESCRIBE MIGRATION" section below for details.
+
+* The ``ALTER CURRENT MIGRATION REJECT PROPOSED`` statement is used to
+  inform the server that the currently proposed statement returned by
+  a prior call to ``DESCRIBE CURRENT MIGRATION AS JSON`` is not acceptable
+  and that the next execution of ``DESCRIBE`` should return an alternative
+  proposal.  If the client rejected all proposed statements, the next
```
tailhook, line 281: How does CLI finds out this? `complete: false` and no `proposed`? Seems to be very hacky way to do that, I would made something more explicit.

Also is there a way to find out what needs to be done? I.e what current diff is. CLI should show user something in this case.

text/1000-migrations.rst
tailhook: Maybe more accent/clarification on this, like "without loosing data that was already committed before applying migration".

text/1000-migrations.rst
```diff
@@ -332,47 +359,88 @@ The returned JSON conforms to the following pseudo-schema::
...
-                    "CREATE PROPERTY number -> int64; }"
-          }],
-          "confidence": 0.4
-        }
-      ]
+      "proposed": {
+        "statements": [{
+          "text": "ALTER TYPE Address " +
+                  "ALTER PROPERTY number " +
+                  "SET TYPE int64 USING \(expr)",
```
tailhook, line 374: By the way, we don't have the description of this variable substitution, do we? Does CLI validate them somehow? (at least lexemes and parenthesis?) Are there constant substitutions where we can validate that variable is of the valid type, or is it just expression? Can there be any other parts of statements except expressions?

text/1000-migrations.rst
elprans: The only valid representation of the diff is DDL, and we've already shown the user all possible variants of it.  I don't think dumping the internal representation is a good idea.

> `complete: false` and no `proposed`? Seems to be very hacky way to do that,

Why?  Looks fairly unambiguous to me.

text/1000-migrations.rst
```diff
@@ -277,18 +303,21 @@ Synopsis::
...
 
 The returned JSON conforms to the following pseudo-schema::
 
     {
+      // Name of the parent migration
+      "parent": "m1...",
+
+      // Whether the confirmed + proposed DDL makes
+      // the migration complete.
+      "complete": {true|false},
```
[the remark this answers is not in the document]
elprans: Sure, we can make it mean that "confirmed" represents the entire migration.

text/1000-migrations.rst
elprans: The CLI should not attempt to interpret expressions or substitutions, treat it as string formatting.  Validation is done on the server.

text/1000-migrations.rst
elprans: OK, will clarify.

text/1000-migrations.rst
elprans: Added a clarification.

elprans:
Is this OK to merge?

PR #19, opened 2020-09-04
Initial Transactions API RFC

tailhook:
Postponed things:
1. Retrying single request (lets tackle it later)
2. `.retry` method on Connection objects

Things not discussed yet:
1. `RetryOptions` class (or is it plain object/keyword arguments)
2. Default options: attempts and delay function

Things to be determined while implementing:
1. Is it possible to make exception class hierarchy or we need to do `is_xxx` method(s)?
2. Whether it's possible to differentiate `EarlyNetworkError`


text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+
+**Introduce** two methods on a connection pool and one on connection:
+
+.. code-block:: typescript
+
+    type TransactionBlock = (Transaction) => Promise<T>;
+
+    interface Pool {
+        retry<T>(block: TransactionBlock, options: RetryOptions): T;
+        raw_transaction<T>(action: TransactionBlock, options?: TransactionOptions): T;
```
elprans, line 119:
```suggestion
        try_transaction<T>(action: TransactionBlock, options?: TransactionOptions): T;
```

?

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+   number is exhausted
+
+The block is introduced by ``connection_pool.retry``, the exact code
+is different for different languages because of language limitations
+and convetions.
+
+Normal transactions that aren't retried are executed with
+``try_transaction`` method.
+
+The ``retry`` function configured by the number of attempts and delay function.
```
elprans, line 85: I suggest using the term "backoff" instead of "delay".

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+Current attempt number N is global. Which means if the last error is a
+deadlock and N is greater than the number of attempts on a deadlock we
+stop retrying and return error (even if previous error was a network
+error).
+
+Delay function by default is ``2^N * 100`` plus random number in range
+``0..100`` microseconds. Where first retry (second attempt) has ``N=1``.
+Technically:
+
+* In JavaScript: ``n => (2**n) * 100 + Math.random()*100``
```
1st1, line 98: TIL: there's `**` operator in JS.

text/1004-transactions-api.rst
1st1: Yes, I like `try` more than `raw`, as the former term plays nicely with `retry`. Also, all JS APIs should use camelCase, so it's

```suggestion
        tryTransaction<T>(action: TransactionBlock, options?: TransactionOptions): T;
```

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+    }
+
+The ``PoolOptions`` object receives ``retryOptions`` field:
+
+.. code-block:: typescript
+
+    interface PoolOptions {
+      // ...
+      retryOptions?: RetryOptions;
+    }
```
1st1, line 202: You should add an important point for both JS and Python APIs: if there's an open `tryTransaction` or a `retry` block, it's illegal to use the `Executor` methods on connections (until the transaction block is closed)

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+
+Exceptions API
+``````````````
+
+Error hierarchy is amended by introducing ``TransientError``, ``NetworkError``
+and ``EarlyNetworkError`` with the following relationships:
+
+.. code-block:: typescript
+
+    class TransientError extends TransactionError {}
```
1st1, line 212: I'd reference the https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt file.  And it looks like some of these exceptions are server-side and some are client side. Please sort them and inherit from appropriate base classes in `errors.txt`.

I'm also not sure why we need `TransientError`. Why is `ExecutionError` (already defined) not sufficient?

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+
+Pool methods for creating a transaction:
+
+.. code-block:: python
+
+   class AsyncIOPool:
+       def retry(
+           options: RetryOptions = None
+       ) -> AsyncIterable[AsyncContextManager[AsyncIOTransaction]]: ...
+       async def raw_transaction(*,
```
1st1, line 237:
```suggestion
       async def try_transaction(*,
```

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+           readonly: bool = None,
+           deferrable: bool = None,
+       ) -> ContextManager[Transaction]: ...
+
+Example usage of ``retry`` on async pool:
+
+.. code-block:: python
+
+    async for tx in db.retry():
+      async with tx:
```
1st1, line 272: I'd also add a quick implementation spec of `retry` in Python. Specifically:

1. `retry()` returns an (async-)iterator which has no methods.
2. Every yielded element is a transaction object, strongly referencing the iterator that created it internally.
3. If the code in the `async with` / `with` block succeeds, the transaction object messages its iterator to stop iteration.

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+where we catch network error before sending a request.
+
+
+Future Work
+===========
+
+Transaction on Specific Connection
+----------------------------------
+
+Do we want and how ``connection.retry()`` should work?
```
1st1, line 384: There's probably a typo here, not sure I can understand the "do we want and how" part

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+
+a. It may only retry on the same connection and fail on disconnect
+b. It may reconnect and replace underlying socket in the connection
+   object and retry
+c. We may only allow transactions on connection pools
```
1st1, line 389: I'm not sure I understand any of these options. Is it about handling a network disconnect during an active transaction? FWIW, to me both (b) and (c) sound outright impractical. I suggest either expanding this section to give the reader more context or just remove it.

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+code into a transaction rather than just a mutation. I.e. retrying a
+single mutable request on a "concurrent update" error must be a
+deliberate decision.
+
+
+Retry Single Queries on Connection Error
+----------------------------------------
+
+To make changing EdgeDB address or restarting EdgeDB work nicely, we
+need to retry simple queries on connection errors too:
```
1st1, line 410: "we need to retry simple queries on connection errors too" this is very questionable (and it looks like it can only apply to the pooling APIs specifically, right?)

IMO, crashing on `ConnectionError` is the only sane option, and those who want to make their code more resilient can create a wrapper API.

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+The idea is that there will be ~1 failure per second. So on local
+instance when testing manually it would hit almost every request (which
+is fine as repeating them shouldn't be prohibitively costly). But under
+a huge load of thousands requests per second, one retry per second
+doesn't influence anything so even for production and/or benchmarks this
+is fine.
+
+I think it should be disabled by an explicit command-line argument like
+``--disable-failure-injection`` but might be tweaked with configuration
+settings?
```
1st1, line 485: I like the idea of injecting failures a lot, but just not as something enabled by default. Fundamentally, RDBMS should make all efforts to make operation succeed safely and nicely. Injecting errors would break code that uses the `tryTransaction` API and systems that are just not created for with high-loads in mind.

Instead I suggest adding a session-level setting so that users can enable failure injection in their staging apps and test frameworks.

text/1004-transactions-api.rst
```diff
@@ -0,0 +1,619 @@
...
+
+.. code-block:: python
+
+    def handler(req, db: edgedb.Pool):
+        do_something(req)
+
+        @db.retry()
+        def my_tx(transaction):
+            let val = transaction.fetch("...")
+            transaction.execute("...", process_value(val))
```
1st1, line 536: Since we do want sync and async APIs look similar, I suggest adding an explicit `my_tx()` call.

1st1 (commented):
This is really great, Paul. I've left a few comments. Let's aim for quick merging and continuing the discussion in GH Discussions. I'll do an editing pass later to fix nits and typos.

text/1004-transactions-api.rst
tailhook: Added

text/1004-transactions-api.rst
tailhook:
```
class InvalidValueError(ExecutionError):
class IntegrityError(ExecutionError):
class TransactionError(ExecutionError):
```
First two must not be retried as far as I understand. There are also a couple of `TransactionError` errors that we must not retry:
```                                                                     
            raise errors.TransactionError('already in explicit transaction')                                                                
        raise errors.TransactionError(f'there is no {name!r} savepoint')                                                                 
            raise errors.TransactionError(f'there is no {name!r} savepoint')                                                                               
            raise errors.TransactionError('already in transaction')                                                                                        
            raise errors.TransactionError('cannot commit: not in transaction') 
```

tailhook: NetworkError is now inherited from ClientError.

text/1004-transactions-api.rst
tailhook: Added

text/1004-transactions-api.rst
tailhook: Improved wording a bit.

text/1004-transactions-api.rst
tailhook: Actually, (c) is what current RFC specifies. I think it's good enough until we have a clear use-case for having `Connection.retry`. If it non-concurrent ad-hoc script. I'd say connection pool with `min_size = 1` is fine. If it's for own connection handling then `try_transaction` is what they should use.

text/1004-transactions-api.rst
tailhook: What's questionable?
```python
def product_index(req):
    return template(self.db.query("SELECT Product{...}"))

def buy_product(req):
    for tx in self.db.retry():
        with tx:
            price = db.query_row("SELECT price := Product.price...")
            db.query_row("INSERT ShopingCart {...}", ...)
```
Now if we restart server, or change address, or do any other networking changes (i.e. in docker you can get a network error even if just changing unrelated container), user can press "buy" button without seeing error page, but can't get a read-only index page without hitting `500 Internal Server Error`.

What we optimize for? Do we recommend read-only transactions? Do we recommend "roll your own wrapper"?

text/1004-transactions-api.rst
tailhook: > I like the idea of injecting failures a lot, but just not as something enabled by default. Fundamentally, RDBMS should make all efforts to make operation succeed safely and nicely. Injecting errors would break code that uses the `tryTransaction` API and systems that are just not created for with high-loads in mind.
> 

This isn't for "high-load" this is about concurrency. Technically any amount of concurrency will get "rare" errors. People using `try_transaction` should be prepared for that.

There are two use cases where it's bad:
1. Handling non-concurrent bulk data insert, i.e. initial preload of the data.
2. Users executing things manually: REPL or IPython

> Instead I suggest adding a session-level setting so that users can enable failure injection in their staging apps and test frameworks.

Well, that's possible but my point that this is not enough. If we recommend `retry` everywhere, users will probably use that, but don't realize how repeating side-effects will create a problem.

Anyway, let's create a separate issue/discussion for this. I think it's not a problem to update RFC later, this section is important to keep in RFC, though.

text/1004-transactions-api.rst
tailhook: If we're adding a call, then it's a different API, I think. Because then it's tempting to move function out of the closure, and ... Anyway this is alternative API to what we're implementing, so I don't think it is important.

1st1:
Let's merge as is and follow up with separate discussions.

PR #24, opened 2020-11-18
Merged discussions to RFC1004

tailhook:
* In particular "Failover and Replication Pre-RFC"
* And "Stateful Connection Configuration Pre-RFC"

### DIFF
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -60,8 +60,9 @@ errors.
 
 `More discussion on this topic <https://github.com/edgedb/edgedb/discussions/1708>`_
 
-This part of API change is also motivated by the fact that PostgreSQL
-sometimes can't apply concurrent transactions and errors out with::
+This part of API change is also motivated by the fact that EdgeDB (as is
+PostgreSQL) sometimes can't apply concurrent transactions and errors out
+with::
 
     Error: could not serialize access due to concurrent update
 
@@ -341,7 +342,7 @@ While removing inherent methods with the same name.
 
 Note: while ``Connection.try_transaction`` block is active,
 ``Executor`` methods are disabled on the connection object itself
-(i.e. they throw ``TransactionActiveError``).
+(i.e. they throw ``TransactionIsActiveError``).
 
 Example of the recommended transaction API:
 
@@ -409,12 +410,12 @@ All network error within connection should be converted into
 ``EarlyNetworkError`` or ``NetworkError``. Former is used in context
 where we catch network error before sending a request.
 
-``TransactionActiveError`` is introduced to signal that
+``TransactionIsActiveError`` is introduced to signal that
 queries can't be executed on the connection object:
 
 .. code-block:: typescript
 
-    class TransactionActiveError extends InterfaceError {}
+    class TransactionIsActiveError extends InterfaceError {}
 
 
 Python Transactions API
@@ -552,7 +553,7 @@ Introduce the abstract classes for queries:
 
 Note: while ``Connection.try_transaction`` block is active,
 ``Executor`` methods are disabled on the connection object itself
-(i.e. they throw ``TransactionActiveError``).
+(i.e. they throw ``TransactionIsActiveError``).
 
 These base classes should be implemented by respective classes:
 
@@ -591,13 +592,13 @@ All network error within connection should be converted into
 ``EarlyNetworkError`` or ``NetworkError``. Former is used in context
 where we catch network error before sending a request.
 
-Additionally ``TransactionActiveError`` is introduced to signal that
+Additionally ``TransactionIsActiveError`` is introduced to signal that
 queries can't be executed on the connection object (i.e. when
 connection object is "borrowed" for the duration of the transaction):
 
 .. code-block:: python
 
-    class TransactionActiveError(InterfaceError):
+    class TransactionIsActiveError(InterfaceError):
         _code = 0x_FF_02_01_04
 
 
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -624,7 +624,7 @@ This RFC introduces a connection parameter:
     await edgedb.connect('inst1', {wait_until_available_sec: 30})
     await edgedb.createPool('inst1', {
         connectOptions: {
-            wait_until_available_sec: 30,
+            waitUntilAvailableMicros: 30,
         }
     })
 
@@ -660,6 +660,9 @@ determines how long individual connect attempt may take. And
 ``wait_until_available`` determines how many such attempts could be
 made (i.e. how many ones fits the time frame).
 
+Default ``wait_until_available`` is 30 seconds or 30000 microseconds
+for JavaScript.
+
 
 EdgeDB Changes
 --------------
@@ -667,9 +670,9 @@ EdgeDB Changes
 To support features above we add two headers to EdgeDB queries:
 
 1. For PrepareComplete_, CommandComplete_ server-side messages:
-   ``SERVER_OPT_HAS_FEATURES``
+   ``COMPILED_QUERY_FEATURES``
 2. For Prepare_, OptimisticExecute_, Execute_, ExecuteScript_
-   client-side messages: ``QUERY_OPT_ALLOW_FEATURES``
+   client-side messages: ``QUERY_ALLOW_FEATURES``
 
 Both contain 64bit bitmap of the following:
 
@@ -681,10 +684,10 @@ Both contain 64bit bitmap of the following:
 5. `PERSISTENT_CONFIG 0b10000` -- server or database config change
 
 In case of ``SERVER_OPT_HAS_FEATURES`` it describes what is actually
-contained in the query. And in case of ``QUERY_OPT_ALLOW_FEATURES``
+contained in the query. And in case of ``QUERY_ALLOW_FEATURES``
 client can specify what of these things are allowed in this query.
 
-Read-only queries are always allowed. When ``QUERY_OPT_ALLOW_FEATURES``
+Read-only queries are always allowed. When ``QUERY_ALLOW_FEATURES``
 is omitted any query is allowed (default). With the bit mask of zero
 only read-only queries are allowed.
 
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -308,7 +308,7 @@ TypeScript API
     }
     class Connection {
         retry<T>(block: TransactionBlock): T;
-        raw_transaction<T>(action: TransactionBlock): T;
+        try_transaction<T>(action: TransactionBlock): T;
     }
 
 Raw connection has only ``try_transaction`` method:
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -93,7 +93,7 @@ Specification
 Connections and Pools
 ---------------------
 
-Instead of exposing a TCP connection to the user we use a set wrapper
+Instead of exposing a TCP connection to the user we use a set of wrapper
 classes:
 
 .. code-block:: python
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -143,7 +143,7 @@ Both ``Pool`` and ``Connection`` have:
 ``ReadOnly*`` counterparts differ from non-read-only ones in two
 important ways:
 
-1. They send read-only flag to the compiler, so compiler rejects queries
+1. They send read-only flag to the database server, which can reject queries
    that modify data. And *may* use replica connection.
 2. Works as type-checker hint that function that received a connection
    does no modifications to the data.
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -246,8 +246,8 @@ a different underlying implementation. E.g. in the future:
    a per-request basis or vice versa make the read-only flag the session
    configuration.
 
-All of these optimizations are seamless for the users. And we expect
-users to trust language bindings to do the optimal thing.
+All of these optimizations would be seamless for the API users,
+who should trust the language bindings to do the optimal thing.
 
 
 Transactions API
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -452,7 +452,7 @@ Pool methods for creating a transaction:
 
    class Pool:
        def retry() -> Iterable[ContextManager[Transaction]]: ...
-       def raw_transaction(
+       def try_transaction(
            isolation: str = None,
            read_only: bool = None,
            deferrable: bool = None,
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -148,7 +148,7 @@ important ways:
 2. Works as type-checker hint that function that received a connection
    does no modifications to the data.
 
-``RawConnection`` class is exposed for users want to do their own
+``RawConnection`` class is exposed for users who want to do their own
 session config management stuff, connection pooling or control of
 reconnection.
 
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -667,24 +667,24 @@ To support features above we add two headers to EdgeDB queries:
 2. For Prepare_, OptimisticExecute_, Execute_, ExecuteScript_
    client-side messages: ``QUERY_OPT_ALLOW_FEATURES``
 
-Both contain the set of strings joined by comma:
+Both contain 32bit bitmap of the following:
 
-1. `modification` -- query is not read-only
-2. `session` -- query contains session config change
-3. `config` -- server or database config change
-4. `ddl` -- query contains DDL
-5. `transaction` -- query contains start/commit/rollback of transaction
-    or savepoint manipulation
+1. `MODIFICATONS      0b00001` -- query is not read-only
+2. `SESSION_CONFIG    0b00010` -- query contains session config change
+3. `TRANSACTION       0b00100` -- query contains start/commit/rollback
+   of transaction or savepoint manipulation
+4. `DDL               0b01000` -- query contains DDL
+5. `PERSISTENT_CONFIG 0b10000` -- server or database config change
 
 In case of ``SERVER_OPT_HAS_FEATURES`` it describes what is actually
 contained in the query. And in case of ``QUERY_OPT_ALLOW_FEATURES``
 client can specify what of these things are allowed in this query.
 
 Read-only queries are always allowed. When ``QUERY_OPT_ALLOW_FEATURES``
-is omitted any query is allowed (default). With the empty string only
-read-only queries are allowed.
+is omitted any query is allowed (default). With the bit mask of zero
+only read-only queries are allowed.
 
-``SERVER_OPT_HAS_FEATURES`` is an empty string for read-only queries (the
+``SERVER_OPT_HAS_FEATURES`` is zero for read-only queries (the
 field is present) as it indicates that query has been analyzed.
 
 The ``SERVER_OPT_HAS_FEATURES`` is needed for the following tasks:
@@ -696,10 +696,11 @@ The ``SERVER_OPT_HAS_FEATURES`` is needed for the following tasks:
 
 By default:
 
-1. Pool supports all except ``transaction,session``
+1. Pool supports all except ``TRANSACTION | SESSION_CONFIG``
 2. Read-only pool and read-only connection support none
-3. Connection warns on ``session`` and ``transaction``
-4. Connection got from pool errors on ``session`` and ``transaction``
+3. Connection warns on ``SESSION_CONFIG`` and ``TRANSACTION``
+4. Connection got from pool errors on ``SESSION_CONFIG`` and
+   ``TRANSACTION``
 5. Everything is allowed in ``RawConnection``
 
 Note: session settings and transactions should be activated using
@@ -1035,13 +1036,13 @@ Disabling Features
 ------------------
 
 We may introduce a pool and/or connection configuration to disable
-``ddl`` and ``config`` features on the requests. And/or disable
-``session`` and ``transaction`` features on connection. This should be
-default for many applications. But since scripting DDL and database
-configuration is also a valid use case and since the risk of misusing
-that is quite small we don't include it into the specification. Also
-disabling ``ddl`` and ``config`` features should be covered by
-permissions.
+``DDL`` and ``PERSISTENT_CONFIG`` features on the requests. And/or
+disable ``SESSION_CONFIG`` and ``TRANSACTION`` features on connection.
+This should be default for many applications. But since scripting DDL
+and database configuration is also a valid use case and since the risk
+of misusing that is quite small we don't include it into the
+specification. Also disabling ``DDL`` and ``PERSISTENT_CONFIG`` features
+should be covered by permissions.
 
 
 External References
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -919,18 +919,20 @@ There are few problems of this approach:
    sub-application might repeat manually an extra repeating automatically
    might make transaction slower and introduce unexpected repeatable side
    effects.
-2. This doesn't help in case of pythonic `with db.transaction()` as we
+2. This doesn't help in case of pythonic ``with db.transaction()`` as we
    allow now.
-3. If we're advising `transaction` on connection object, reconnecting on
-   network failures would be an issue
+3. If we're advising ``transaction`` on connection object, reconnecting
+   on network failures would be an issue
 
 
 Separate Retry API
-------------------------------------------------------
+------------------
 
-The problem with this approach is that it hard to teach using ``retry``
-when raw transactions "work on my laptop". However, this is somewhat
-alleviated by failure injection.
+The problem of adding a separate transaction retry API while keeping the
+current ``Connection.transaction()`` family of methods is that it will
+be hard to educate the users that they should almost always use the
+former.  Most of the time transaction serialization and network errors
+are uncommon during the development and only surface when in production.
 
 
 Retry All Single Queries
--- a/text/1004-transactions-api.rst
+++ b/text/1004-transactions-api.rst
@@ -19,7 +19,7 @@ This RFC describes client API that is:
 1. Resilient against network errors and server failover (i.e. when server
    changes IP address) both within a single transactions and for
    non-transactional queries.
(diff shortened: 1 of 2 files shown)


text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+field is present) as it indicates that query has been analyzed.
 
+The ``SERVER_OPT_HAS_FEATURES`` is needed for the following tasks:
 
-Retry Single Queries on Connection Error
-----------------------------------------
+1. Retry standalone (non-transactional) read-only queries
+2. Warn when features are used in inapropriate context (warn when
+   session modification queries are sent on non-raw connection, which
+   means they can be lost at any point due to reconnect)
```
1st1, line 704: Hm, how will the server know if the connection is raw or non-raw? IMO it shouldn't care, it's entirely up to the client.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
-single mutable request on a "concurrent update" error must be a
-deliberate decision.
+``SERVER_OPT_HAS_FEATURES`` is an empty string for read-only queries (the
+field is present) as it indicates that query has been analyzed.
 
+The ``SERVER_OPT_HAS_FEATURES`` is needed for the following tasks:
 
-Retry Single Queries on Connection Error
-----------------------------------------
+1. Retry standalone (non-transactional) read-only queries
```
1st1, line 701: You mean that the server would retry some of the client's queries on its own? I wouldn't like that.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+===========
 
-To make changing EdgeDB address or restarting EdgeDB work nicely, we
-need to retry simple queries on connection errors too:
+
+More Configuration
+------------------
+
+Setting tracing/debugging metadata for queries may be implemented
+in the same way:
```
1st1, line 737:
```suggestion
in the same way in the future:
```

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
-4. Or vice versa add ``pool.read_only_query()`` method
-5. Add ``pool.read_only().query()``
-6. Add ``pool.with_options({readonly: true}).query()``
+   def read_only(self, primary=False, max_replica_lag=10): ...
 
-Note: methods (2-6) are also helpful for working with primary/replica
-installations. But probably only last two would allow full power, as they allow
-``pool.read_only(primary=true)`` (i.e. in case you need read-only transaction
-that can't go to a replica).
+Perhaps this should be encapsulated into replica options:
```
1st1, line 770: Let's not use "perhaps" here -- we need to decide now or leave this for future RFCs

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+.. code-block:: python
+
+   def read_only(self, primary=False, replicas: ReplicaOptions): ...
 
 
 Learning Curve
 ==============
 
-This complicates learning curve, but:
+The ``retry`` method complicates learning curve curve, but:
```
1st1, line 780:
```suggestion
The ``retry`` method complicates the learning curve, but:
```

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
 
 1. This is already a problem in Postgres, there a lot of people who
-   ignore the issue for pet projects and a lot of startups having "normal"
-   rate or 500 errors at any point of time, while it's preventable.
+   ignore the issue for pet projects and a lot of startups having
+   "normal" rate or 500 errors at any point of time, while it's
+   preventable.
```
1st1, line 785: "1. Letting the application to error out instead of automatically retrying certain transactions is a wildly known, yet an entirely preventable problem."

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
 2. Repeating transactions would make less incentive to keep transactions
-   open for a long time (e.g. while accessing slow network resources like
-   external API), which is a problem of itself.
+   open for a long time (e.g. while accessing slow network resources
+   like external API), which is a problem of itself.
```
1st1, line 788: "2. The design of the proposed `retry()` method emphasizes that the code block might be executed more than one time, suggesting to the user to factor out slow blocking code, like making API calls over network. This ensures that DB transactions would not be open longer than it is necessary."

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
 3. Even if we never have failed concurrent updates we would want
    seamless reconnect on connection failures (i.e. server restart,
    primary/replica change, etc.)
```
1st1, line 791: "3. The API adds resilience against not only errors related to failing to serialize concurrent updates, but also against possible network errors caused by the DB server restart, primary/replica changes, etc."

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
 4. To make learning curves shorter I think we should intentionally
-   inject failures. This is needed so that users quickly find out that side
-   effects of their transactions are in effect several times.
+   inject failures. This is needed so that users quickly find out that
+   side effects of their transactions are in effect several times.
```
1st1, line 794: "4. In the future we will consider adding a connection setting to inject failures into transactions to help users discover incorrect transactional code."

@tailhook @elprans I think we shouldn't do this before 1.0 as I'm not sure how it might impact benchmarks, production code, and the overall DX of early adopters. I think that our API is already going to be structured in the right way, and adding failure injection is entirely optional at this point.

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
    network failures would be an issue
 
 
-Keep `transaction` as but Add a Helper for Retrying
----------------------------------------------------
+Keep `transaction` as is but Add a Helper for Retrying
```
1st1, line 948:
```suggestion
Separate Retry API
```

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
...
+------------------------------------------------------
 
-The problem with this approach is that it hard to teach using `retry`
+The problem with this approach is that it hard to teach using ``retry``
```
1st1, line 951: "The problem of adding a separate *transaction retry API* while keeping the current `Connection.transaction()` family of methods is that it will be hard to educate the users that they should almost always use the former. Most of the time transaction serialization and network errors are uncommon during the development and only surface when in production."

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
...
 
 
+Retry All Single Queries
+------------------------
+
+This specification describes that read-only non-transactional queries
+should be retried automatically.
+
+While it's tempting to retry all ``pool.query`` and ``pool.execute``
+calls (even modifications), it **gives the false sense of security**: no
```
1st1, line 963:
```suggestion
calls (even those that execute queries with side effects), it **gives the false sense of security**: no
```

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
...
+"concurrent update" issues are seen.  But it's better to see such error
+and turn the whole block of code into a transaction rather than just a
+mutation. I.e. retrying a single mutable request on a "concurrent
+update" error must be a deliberate decision.
```
1st1, line 967: This whole paragraph is very hard to parse. I suggest to reword it to:

"Automatically retrying `query()`-like calls that execute queries with side effects is not always safe. Therefore our recommendation to users would be to use the new `retry()` API when they know that the query would be safe to repeat."

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
...
+
+
+Is ``with_modifications`` Needed?
+---------------------------------
+
+It's intuitive that databases are mutable by default. But there is a
+large class of applications that are mostly read-only and must have
+limited and easy to find places having mutations. For those apps it's
+better to have read-only connection by default and use a pattern like
+this for writes:
```
1st1, line 977: The RFC starts with stating and showing examples of `edgedb.connect()` returning a mutable connection. So I assume that this section advocates for adding a way to convert a read-only connection (obtained via `edgedb.connect().read_only()`) to a mutating one?  I'm confused.

text/1004-transactions-api.rst
```diff
@@ -651,16 +945,129 @@ There are few problems of this approach:
...
+configuration anyway. So it deviates from the usual behavior of context
+manager anyway.
+
+
+Disabling Features
+------------------
+
+We may introduce a pool and/or connection configuration to disable
+``ddl`` and ``config`` features on the requests. And/or disable
+``session`` and ``transaction`` features on connection. This should be
```
1st1, line 1058: I like the disabling idea. I'm not sure if it should be the default configuration though, we can just teach that it's a good pattern to narrow down the connection functionality to what your app needs.

1st1 (commented):
Wow. Overall this looks really close. Thanks for championing this and for spending time to convince us that these ideas are the best path forward.

I've left some comments with questions and edit suggestions.  I can do a heavy editing pass once this is merged.

text/1004-transactions-api.rst
elprans: I also think that proper API design, documentation and tutorials would be sufficient to guide users toward the correct set of idioms.  Intentional failure injection might be an optional feature.

text/1004-transactions-api.rst
elprans: Agree.  I think we should optimize for simplicity at a glance, in which case lengthy connection "setups" in simple tutorials would scare users away.

elprans (commented):
Great work, @tailhook!

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
-tempting to add a method on connection object itself:
+This RFC introduces a connection parameter:
 
 .. code-block:: python
 
-    with pool.acquire() as conn:
-        for tx in conn.retry():
-            with tx:
-                tx.query(..)
+    edgedb.connect('inst1', wait_until_available_sec=30)
```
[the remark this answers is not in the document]
1st1: Hard and solid -1 from me. One library shouldn't fight the language for unclear benefits.

text/1004-transactions-api.rst
```diff
@@ -68,6 +89,170 @@ to handle:
...
+* Setting session configuration
+* Transaction options
+
+Here are method signatures:
+
+.. code-block:: python
+
+    class Connection:
+        ...
+        def read_only(self,
```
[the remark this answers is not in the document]
1st1: I think `as_*` is more common for this kind of API.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
 
+In case of ``SERVER_OPT_HAS_FEATURES`` it describes what is actually
+contained in the query. And in case of ``QUERY_OPT_ALLOW_FEATURES``
+client can specify what of these things are allowed in this query.
 
-Retry Single Queries
---------------------
+Read-only queries are always allowed. When ``QUERY_OPT_ALLOW_FEATURES``
+is omitted any query is allowed (default). With the empty string only
+read-only queries are allowed.
```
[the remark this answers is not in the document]
1st1: I don't understand why we need this complexity.  IMO instead of "modification" header flag we should have "readonly". If it's set the query fails to prepare unless it's readonly. This is way easier to understand and document.

text/1004-transactions-api.rst
1st1: Yeah. Paul, we've argued over this a lot already, and I see no end to this. So I propose we do this instead: we don't implement `with_modifications()` right now. We can always add this later, it's really not a critical feature and we can add it at any time when we get some user feedback.

I'd really like this RFC to only implement essential and **non-controversial** APIs, where non-controversial means unanimous support of you, Elvis, and I.

text/1004-transactions-api.rst
```diff
@@ -68,6 +89,170 @@ to handle:
...
+1. They send read-only flag to the compiler, so compiler rejects queries
+   that modify data. And *may* use replica connection.
+2. Works as type-checker hint that function that received a connection
+   does no modifications to the data.
+
+``RawConnection`` class is exposed for users want to do their own
+session config management stuff, connection pooling or control of
+reconnection.
+
+Note: this section contains only sync Python example, async Python and
```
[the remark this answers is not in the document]
tailhook: Oh, well, forgot that entirely. My idea was that `AsyncIO` prefixes and `async` in signatures don't add any value, but...

text/1004-transactions-api.rst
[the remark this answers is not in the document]
tailhook: Then we have to change all other methods. But `as_session_config` is not the same as `with_session_config`. So let's make `with_read_only` then?

text/1004-transactions-api.rst
```diff
@@ -68,6 +89,170 @@ to handle:
...
+
+.. code-block:: python
+
+    conn = edgedb.connect()
+    read_only_conn = conn.read_only()
+    conn.execute("INSERT User { ... }", ...)
+    print(read_only_conn.query("SELECT User"))
+    read_only_conn.execute("INSERT User { .. }", ...)  # throws an error
+    conn.read_only(allow_upgrade=True) \
+        .with_modifications().execute("INSERT User { ... }", ...)  # good
```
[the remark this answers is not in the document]
tailhook: Well, I still think this is very useful pattern that I would use in **every** my project. But if we have to cut it to do at least one step forward, okay.

text/1004-transactions-api.rst
```diff
@@ -224,7 +416,7 @@ All network error within connection should be converted into
 ``EarlyNetworkError`` or ``NetworkError``. Former is used in context
 where we catch network error before sending a request.
 
-Additionally ``TransactionActiveError`` is introduced to signal that
+``TransactionActiveError`` is introduced to signal that
 queries can't be executed on the connection object:
```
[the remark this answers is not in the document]
tailhook: Added explanation

text/1004-transactions-api.rst
```diff
@@ -224,7 +416,7 @@ All network error within connection should be converted into
 ``EarlyNetworkError`` or ``NetworkError``. Former is used in context
 where we catch network error before sending a request.
 
-Additionally ``TransactionActiveError`` is introduced to signal that
+``TransactionActiveError`` is introduced to signal that
```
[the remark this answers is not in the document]
tailhook: `TransactionIsActiveError`? 

`ActiveTransactionError` -- reads as error of active transaction (i.e. some base for a whole class of errors).

Also added clarification below.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
-Future Work
-===========
+Server Availability Timeout
+---------------------------
 
-Transaction on Specific Connection
-----------------------------------
+Previously, when TCP connect gets "connection refused" error or when
+timeout happens on handshake both connection and pool API would crash
+the application. This is weird when initially starting an application
```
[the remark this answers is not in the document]
tailhook: Changed to "can be inconvenient". This isn't particularly confusing.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+3. Connection reset, connection aborted, connection refused (server is
+   restarting or not ready yet)
+4. Timeout happened during connect or authentication
+
+All other errors are propagated immediately.
+
+This timeout is different from ``connect_timeout``. Connect timeout
+determines how long individual connect attempt may take. And
+``wait_until_available`` determines how many such attempts could be
+made (i.e. how many ones fits the time frame).
```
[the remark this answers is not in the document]
tailhook: Previously we discussed that at least with @1st1 and he was for this change.

Here are my points:
1. For nice `docker-compose up` we need `wait_until_available` = 30 sec
2. Current python timeout is 60 sec. So by introducing `wait_until_available` 30 sec we don't make application hanging longer than what was there. In particular, I propose to lower `connect_timeout` to 10 seconds (for a single connection attempt). With connection pooling (ports RFC) this should work well enough.
3. If default is `0` nobody would ever discover `wait_until_available` and will put their own hacks for running locally, like they do with postgres. Unless we put this setting in every tutorial (which we definitely don't want for conciseness).
4. We can't crash app on every `query()` that got reconnected and caught `ConnectionRefused` (and get app restarted). So it's also a bunch of 500 errors on EdgeDB restart or failover.

*(the case when app and database starts simultaneously with the database on kubernetes is so much rare that we don't care about that use case)*

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+EdgeDB Changes
+--------------
+
+To support features above we add two headers to EdgeDB queries:
 
-While this covers "concurent update" and "deadlock" errors, with the
-current implementation it can't reconnect when when connection was
-interrupted while transaction is active. There are few ways it can work:
+1. For PrepareComplete_, CommandComplete_ server-side messages:
+   ``SERVER_OPT_HAS_FEATURES``
```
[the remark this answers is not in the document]
tailhook: It's kinda just symmetry with ``QUERY_OPT``. ``SERVER_HAS_FEATURES`` reads bad. But better name is appreciated.

text/1004-transactions-api.rst
tailhook: Not server, client.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+1. `modification` -- query is not read-only
+2. `session` -- query contains session config change
+3. `config` -- server or database config change
+4. `ddl` -- query contains DDL
+5. `transaction` -- query contains start/commit/rollback of transaction
+    or savepoint manipulation
 
+In case of ``SERVER_OPT_HAS_FEATURES`` it describes what is actually
+contained in the query. And in case of ``QUERY_OPT_ALLOW_FEATURES``
+client can specify what of these things are allowed in this query.
```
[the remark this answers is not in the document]
tailhook: Not sure I understand. It's in the next paragraph.

text/1004-transactions-api.rst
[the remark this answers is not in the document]
tailhook: This is because:
```
conn.execute("""
START TRANSACTION;
UPDATE User SET {...}
COMMIT
""")
```
Would return `transaction,modifications` flag.

With @1st1 proposal `.execute("START TRANSACTION")` would return `transaction,readonly` features and `.execute("START TRANSACTION; UPDATE..."` would return `transaction`.

I.e. purely additive features are better at least for `HAS_FEATURES`, but is not bad for `ALLOW` too.

I we go with bitmask read-only queries == '0' is also good enough.

text/1004-transactions-api.rst
tailhook: Here is the strawman (but all of these are implementation details):
```
class RawConnection:
    def _query_with_options(self, options, query, *args, **kw): ...
    def query(self, query, *args, **kw):
        self._query_with_options({'allow': '*'}, query, *args, **kw)

class Connection:
   def query(self, query, *args, **kw):
       self._raw._query_with_options({'warn': ['transaction','session_config']}, query, *args, **kw)

class ReadonlyConnection:
   def query(self, query, *args, **kw):
       self._raw._query_with_options({'allow': []}, query, *args, **kw)

class PoolConnectionProxy:
   def query(self, query, *args, **kw):
       self._raw._query_with_options({'deny': ['transaction','session_config']}, query, *args, **kw)

class Transaction:
    def __enter__(self):
          self._conn._raw._query_with_options({'allow': ['transaction']}, "START TRANSACTION")
    def query(self, *args, **kw):
          self._conn._raw._query_with_options({'allow': ['modifications', 'ddl']}, query, *args, **kw)
````
(these allow/deny are to show the concept, we're probably going to have a constant with the bit mask already prepared for the specific use case for efficiency)

File: text/1004-transactions-api.rst [hunk 27]
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+Future Work
+===========
 
-To make changing EdgeDB address or restarting EdgeDB work nicely, we
-need to retry simple queries on connection errors too:
+
+More Configuration
+------------------
+
+Setting tracing/debugging metadata for queries may be implemented
```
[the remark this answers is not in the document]
tailhook: That is a future work. Tags are essentially a specific way to organize metadata. Also OpenTracing is deprecated, and OpenTelemetry should be used instead.

Also `with_tags` has little meaning. `with_telemetry_tags` or `with_tracing_tags` looks better, but let's not bikeshed about this here.

text/1004-transactions-api.rst
tailhook: This is a "Future Work" section. What's wrong in "perhaps" here? This just gives more context to the current decisions but don't have to be clearly defined in this RFC.

text/1004-transactions-api.rst
tailhook: This is where we constantly disagree. Documentation and tutorials are not enough for the same reason for which we use unit tests to test code. As we now basically know the distinction between workstation installations (created by `edgedb server init`) and production (roughly -- everything else), it's possible to have server itself without failure injection by default, but `server init` configuring it in a way that enables this by default. Also that section describes a way of failure injection that doesn't impact benchmarks in any significant way.

But anyway, this RFC is already hanging for too long, so I drop the "failure injection" section in favor of later RFC.

text/1004-transactions-api.rst
tailhook: This is not enough in my opinion. Added an example for clarification.

text/1004-transactions-api.rst
tailhook: > So I assume that this section advocates for adding a way to convert a read-only connection (obtained via edgedb.connect().read_only()) to a mutating one?

Yes, in the top of the application, you either do (assuming `aiohttp`):
```python
app['edgedb'] = await create_async_pool('name')
```
Or you do:
```python
app['edgedb'] = (await create_async_pool('name')).read_only(allow_upgrade=True)
```
In the latter case, you prevent arbitrary functions execute side-effects unless run within `with_modification().retry()` transaction and transaction is passed to the function directly.

> In my mind, the main feature and premise of read_only() is that it allows better scalability by allowing connections to standby replicas.

Well:
1. We don't implement that scalability just yet anyway
2. With regards of this RFC the original premise of read-only was retrying queries automatically. But well, it's solved by `HAS_FEATURES` so is not strictly needed
3. I don't see anything in `with_modifications` as controversial (except the name). Controversial would be official recommendation of this, but not the support. And in previous discussion I think it was agreed as "yes we can support this pattern".

Anyway I'm dropping the function and move this section to the Future Work, so that we can finally move forward.

text/1004-transactions-api.rst
tailhook: Looks like you agree with what is written in the RFC. Resolving the conversation.

text/1004-transactions-api.rst
```diff
@@ -39,7 +60,7 @@ errors.
 
 `More discussion on this topic <https://github.com/edgedb/edgedb/discussions/1708>`_
 
-Second part of the RFC is mainly motiveated by the fact that PostgreSQL
+This part of API change is also motivated by the fact that PostgreSQL
```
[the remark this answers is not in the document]
tailhook: My point was that unprepared reader can read this and say that "edgedb is less stable and throws more errors than postgres and that all this transaction repeating thing is a workaround to handle this", which isn't true. Another backend could tackle this issue differently (i.e. locking everything). And we generally don't hide that we're using postgres as a backend.

With all that said should I replace Postgres to EdgeDB?

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
-b. It may reconnect and replace underlying socket in the connection
-   object and retry
-c. We may only allow transactions on connection pools (current
-   specification)
+Both contain the set of strings joined by comma:
 
-This can be postponed to later RFC when we know what are use cases of
-it.
+1. `modification` -- query is not read-only
+2. `session` -- query contains session config change
```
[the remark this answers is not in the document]
tailhook: Changed and also use bitmap, so names matter much less.

text/1004-transactions-api.rst
```diff
@@ -403,107 +601,197 @@ where we catch network error before sending a request.
...
+3. `config` -- server or database config change
+4. `ddl` -- query contains DDL
+5. `transaction` -- query contains start/commit/rollback of transaction
+    or savepoint manipulation
```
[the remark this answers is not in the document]
tailhook: I don't think that performance here would be noticeable, but bitmap sounds good anyway. and using all ones to "allow everything" is good too.

text/1004-transactions-api.rst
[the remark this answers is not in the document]
tailhook: Fixed

text/1004-transactions-api.rst
```diff
@@ -667,24 +667,24 @@ To support features above we add two headers to EdgeDB queries:
 2. For Prepare_, OptimisticExecute_, Execute_, ExecuteScript_
    client-side messages: ``QUERY_OPT_ALLOW_FEATURES``
 
-Both contain the set of strings joined by comma:
+Both contain 32bit bitmap of the following:
```
elprans: Let's do 64 bits.  No reason to conserve space here.

text/1004-transactions-api.rst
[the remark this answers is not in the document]
elprans: I agree with @tailhook.  It's easier to reason when each bit _extends_ capabilities consistently as opposed to either restricting or extending them.

text/1004-transactions-api.rst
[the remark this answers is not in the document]
elprans: I think writing it as "EdgeDB (as is PostgreSQL)" strikes a good balance between your and Yury's concern.

PR #27, opened 2021-01-04
Second attempt at the Consistent CLI Design RFC

ambv:
Old body kept for posterity. The amount of change is big enough it's easier to read this only looking at the new content though.

### DIFF
--- a/text/1003-cli-naming.rst
+++ b/text/1003-cli-naming.rst
@@ -1,7 +1,7 @@
 ::
 
     Status: Draft
-    Type: Feature
+    Type: Guideline
     Created: 2020-07-02
     RFC PR: `edgedb/rfcs#0014 <https://github.com/edgedb/rfcs/pull/14>`_
 
@@ -10,7 +10,7 @@
 RFC 1003: Consistent CLI Design
 ===============================
 
-This RFC defines the naming scheme for EdgeDB command line tools
+This RFC discusses the naming scheme for EdgeDB command line tools
 and their shortcuts in REPL.  The RFC also documents which command line tools
 can be exposed in REPL and how, as well as how various REPL options should be
 specified.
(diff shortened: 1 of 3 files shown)


text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
-    list-object-types
-    list-ports
-    list-roles
-    list-scalar-types
-    query
-    restore
-    server
 
-and::
+Guidelines for future CLI commands
```
ambv, line 27: Those are guidelines summed up from our discusion over Zoom about this topic.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
 
-Furthermore, RFC 1000 adds more top-level subcommands, such as:
+**Concern**: With more and more functionality exposed to the CLI, the
+``<action>-<category>`` approach has the potential to balloon the output
+of ``edgedb -h`` to more than 100 lines, making it harder to read.
 
-* ``edgedb migrate``,
-* ``edgedb create-migration``,
-* ``edgedb show-status``.
+**Decision**: a long but flat ``--help`` provides better discoverability
```
ambv, line 62: ACTIONABLE ITEM 1: invoke a pager on long --help.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+of status it will show. It could be showing status of the current migration,
+or of the currently running server, or of how many active connections the
+server has at the moment. Renaming it to ``edgedb show-migration-status``
+would make it longer to type and overall very verbose. With
+``<group> <command>`` scheme the solution is obvious:
+``edgedb migration status``.
 
-4. The ``<action>-<category>`` scheme does not allow category-specific
-   help sections.
+**Decision**: ``edgedb show-status`` should be renamed to
```
ambv, line 103: ACTIONABLE ITEM 2: rename `show-status` to `migration-status`.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+servers.
 
+**Decision**: the CLI is an unlikely place for a user to learn about new
+complex functionality like the migration subsystem. Better suited avenues
+for this are the online documentation and/or ``man`` pages, or better yet,
+``tldr``.
 
-Overview
-========
+To address discoverability of related sub-commands, top-level listing of
```
ambv, line 129: (maybe) ACTIONABLE ITEM 3: re-order sub-command listing by functionality.

This is my personal suggestion to address discoverability of related commands that are not listed one by one because they're not adjacent in the alphabetical order. If this is hard to implement or deemed not needed, I'll edit this out.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
 
-The RFC proposes to adopt the ``<group> <command>`` naming scheme for
-the ``edgedb`` CLI.  Here's an outline of current and future ``edgedb``
-subcommands:
+``configure`` uses ``<group> <command>`` whereas ``list-`` is flat
+------------------------------------------------------------------
+
+**Concern**: ``configure`` is inconsistent with ``list-``.
+
+**Decision**: It should be replaced by ``edgedb insert-config``,
```
ambv, line 137: ACTIONABLE ITEM 4: rename `configure`.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+**Decision**: the CLI should keep listings of entities that it can
+directly manipulate with other sub-commands. This includes:
+
+* ``list-databases`` (through ``create-database``);
+
+* ``list-ports`` (through ``configure``); and
+
+* ``list-roles`` (through ``alter-role``).
+
+Other listings should be removed as they are redundant with REPL
```
ambv, line 174: ACTIONABLE ITEM 5: remove most list-* commands.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+functionality. Instead, the following form should be enabled::
+
+  $ echo "\list-object-types" | edgedb
```
ambv, line 177: (maybe) ACTIONABLE ITEM 6: allow piping `\command` to `edgedb`.

@tailhook warned in #14 that this opens a can of worms. From what I read on that thread, the concern was only about this syntax being less convenient. While that's true, thrashing `--help` with the current long list of `list-*` sub-commands is also inconvenient. So that's a balancing act.

Also, `sqlite3` and possibly other CLIs allow this.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+The REPL help for ``\\`` commands is overloaded
+-----------------------------------------------
+
+**Concern**: It is convenient to expose the EdgeDB CLI commands directly in
+REPL. This can greatly simplify administrative tasks when the DB
+administrator has full DB admin rights but yet can't access the server shell.
+Sadly, the presence of those commands under the ``\\`` prefix mixes them
+up with the REPL-specific escape commands like ``\last-error``.
+
+**Decision**: Expose CLI-specific commands in the REPL via the ``!`` prefix,
```
ambv, line 190: (maybe) ACTIONABLE ITEM 7: allow CLI access through a `!` prefix

@1st1  proposed this and I like it. It speaks to my Vim and IPython-trained brain. @tailhook doesn't like it because it will make finding the right option take twice as long. This needs to be decided.

text/1003-cli-naming.rst
```diff
@@ -19,110 +19,238 @@ specified.
...
+                                    `edgedb server init` to connect to
+                                    (overrides host and port)
+  -u, --user <user>                 User name of the EdgeDB user
+
+This sometimes creates unreadable commands through inconvenient syntax,
+for instance::
+
+  $ edgedb -d tutorial dump tutorial.edgedb
+
+**Decision**: Passing *connection* options should be allowed anywhere
```
ambv, line 234: ACTIONABLE ITEM 8: allow passing connection options inside sub-commands

IIUC, this is currently impossible in `clap`. So we might need to drop it unless @1st1 decides it's very important. I find it pragmatic so I'd be +1 to have it if at all possible.

text/1003-cli-naming.rst
1st1: +1 for the pager.

Also, I remember we discussed this on one of the calls: how about we show short help for `-h` and long, possibly paginated help for `--help`? That's what `git` does, btw, and I quite like it.

text/1003-cli-naming.rst
1st1: +1.

text/1003-cli-naming.rst
1st1: Big +1.

text/1003-cli-naming.rst
1st1: I'm OK with this, although I'm also OK to keep it as is. @elprans ?

text/1003-cli-naming.rst
1st1: +1. At least until there's a clear user request/reason to have them exposed in CLI and not just in REPL.

text/1003-cli-naming.rst
1st1: -1 to allow piping backslashed commands. IMO this should be requested by the users before we think about implementing this.

text/1003-cli-naming.rst
1st1: I'm biased here, but I still do like the idea even after a few months after proposing it. IMO, repl-specific commands like `\lt` are different from general-purpose shell commands like `dump`.

text/1003-cli-naming.rst
1st1: I'd love this to be possible and I'm wondering how hard it would be to make it so.

1st1 (approved):
IMO this is ready for merging (and answering actionable items in follow up PRs), although let's wait a day or so for @tailhook, @elprans, and maybe others from @edgedb/engineering to chime in.

text/1003-cli-naming.rst
tailhook: It requires a bit of manual work, but we can do that.

As an alternative we can do something like:
```
SUBCOMMANDS:
   <alphabetical link>
   
FURTHER HELP:
   edgedb help migrations  -- more help on how to maintain migrations (not just list of commands)
   edgedb help server -- installing server and maintaining local installation (in fact this already works)
   edgedb help something-else  -- more help on how to maintain migrations
```
I'm not sure what are groups and topics are though.

text/1003-cli-naming.rst
tailhook: We can hide them from the help for brewity, if you wish. But I don't see the reason to remove them. `edgedb list-object-types` can be used in CI to find out what's going on. Also `edgedb list-object-types | cat` emits tab-separated data that can be used for scripting.

text/1003-cli-naming.rst
tailhook: Even if they are different, we will have to do:
```
some-db> \dump my.dump
No such command `dump`.
Did you mean `!dump`?
> !list-databases
Do such command `list-databases`
Did you mean `\list-databases`?
````
Which already smells like we have to parse them equally anyway, just create unnecessary hassle for users.

I still don't see what's the problem with unifying them. In this case, this question doesn't even exist.

File: text/1003-cli-naming.rst (hunk 9 above)
Comment (tailhook) [reply text/1003-cli-naming.rst:234]: Both of you ignore what [I've already said](https://github.com/edgedb/rfcs/pull/14#discussion_r449470218):
1. We have conflicting options already (see the link).
2. `dump --all --database something` is weird
3. How would it work with `configure`, can you specify options on every level?
4. How would this work with `server`, can you specify options after `server`?
5. Do they work in slash commands `\dump --database xxx` ?
6. Someone have to inspect if there are any other conflicts than what I've listed. And only after we figure out what to do with all of them we can +1 on this item
7. Also what to do with `--tab-separated` and similar output options? Do we keep them on the top level? Do we propagate them to all the same commands? Do we propagate them only to commands that it makes sense to (e.g. `list-*` and `query` but not `dump/restore`

(Note: if answer on (3) or (4) or (5) is negative, which I would assume so at least for (4), clap's support of option propagation will not help us anyway, i.e. we have to do it manually).

File: text/1003-cli-naming.rst [hunk 10]
```diff
@@ -19,110 +19,238 @@ specified.
...
+
+  >>> !dump db tutorial
+  done
+
+This way the internal REPL help system is not overloaded with rarely
+needed help on CLI commands and would only show the list of convenient ``\``
+commands with a hint that ``!help`` or ``!h`` can be used to list
+all CLI options.
+
+Name-spacing ``\help`` and ``!help`` is good for the usability, because
```
tailhook, line 207: We can probably have `\help`, which hides infrequent commands like `dump` and `\help --all`.

text/1003-cli-naming.rst
1st1: > It requires a bit of manual work, but we can do that.

Awesome.

> As an alternative we can do something like:

I think hand-sorting help can make it significantly easier to parse for humans. So I'd try that.

text/1003-cli-naming.rst
1st1: > edgedb list-object-types can be used in CI to find out what's going on.

Or you could just pipe an introspection query to the repl. But I hear you.

Getting back to the topic of `-h` and `--help` being slightly different -- maybe we omit them from `-h` but show everything in `--help`?

Also, @elprans, please join.

text/1003-cli-naming.rst
1st1: > Both of you ignore what I've already said:

Sorry, it's been a while since I last saw it. I remembered it as soon as I clicked on the link though.

> We have conflicting options already (see the link).

Maybe we rename those conflicting options? `create role` and `alter role` can use `--set-password` and `--set-password-from-stdin` which actually are more readable and don't conflict with connection options.

> `dump --all`

I think we opted for having a separate `dump-all` command, as it's output and functionality is quite different from the basic `dump`.

> How would it work with configure, can you specify options on every level?

It's two levels. `edgedb [connection config] <command> ...` or `edgedb <command> [connection config] ...`. No need to use it for further nested subcommands.

The reason for that is that at least I view `edgedb dump` as an atomic command on its own, so it feels _really weird_ to me to say `edgedb -d db dump` and not `edgedb dump -d db`.

> Do they work in slash commands \dump --database xxx ?

Why not? Some slash commands already have a bunch of `--options`.

> Someone have to inspect if there are any other conflicts than what I've listed. And only after we figure out what to do with all of them we can +1 on this item

@ambv can you take a quick look?

> Also what to do with --tab-separated and similar output options? Do we keep them on the top level? Do we propagate them to all the same commands? Do we propagate them only to commands that it makes sense to (e.g. list-* and query but not dump/restore

I only care about connection options specifically (database name, host, etc).  IMO things like `--tab-separated` should follow the command, so it's going to be `edgedb <subcommand> --tab-separated`, if it supports it.

text/1003-cli-naming.rst
1st1: Alright, I'm now convinced it's indeed a bad idea. We'll introduce confusion where its not really necessary. -1 from me too.

text/1003-cli-naming.rst
tailhook: > Why not? Some slash commands already have a bunch of --options.

A couple of points:
1. I wanted to avoid:
    ```
    some_database> \list-object-types -d other_database
    ```
2. And also avoid listing a bunch of options in `\list-object-types --help` (because there are two useful options and a much more connection options)
3. Options like `--password` are weird here.
4. Using `-d database` where `--password` were specified on command-line is also weird (users might expect password to be reused? Or that password is asked again?)

So generally because this is how repl works: all commands are executed on the same connection.

> I think we opted for having a separate dump-all command, as it's output and functionality is quite different from the basic dump.

I don't see it mentioned anywhere. @ambv ?

But it also doesn't solve the issue:
```
edgedb dump -u my_user -d my_database file.dump
```
Looks perfectly fine. But:
```
edgedb dump-all -u my_user -d my_database file.dump
```
Looks weird, but *may be useful* if there the `edgedb` database was dropped.

> I only care about connection options specifically (database name, host, etc). IMO things like --tab-separated should follow the command, so it's going to be edgedb <subcommand> --tab-separated, if it supports it.

I'm asking because they are also top-level options, that in turn mostly because they are options for non-interactive scripting (and that mode invoked without any subcommand). And they can easily be applicable to almost all the same commands as connection options (sans dump/restore), i.e. `edgedb --json migration-log`, mutation commands can also output status in JSON like `edgedb --json drop-role` (although the use case for latter is not clear).

So the general approach is that we don't propagate all top-level options. We just hand-picking some of them to some commands?

1st1: >> I think we opted for having a separate dump-all command, as it's output and functionality is quite different from the basic dump.
>
> I don't see it mentioned anywhere. @ambv ?

I think we discussed it on one of our calls. Perhaps @elprans remembers? But I'm strongly in favor of making dumpall a separate CLI command.

1st1: > So the general approach is that we don't propagate all top-level options. We just hand-picking some of them to some commands?

Yeah, that's what I'm thinking we should do. Generally, we'll have some options that are common and can be applied in subcommands too, and some will be REPL-specific (and it should be an error to try to use a REPL-specific option on a subcommand).

tailhook: > > Looks weird, but may be useful if there the edgedb database was dropped.
> 
> I'm not sure I follow that example. You still need to connect to the DB to dump all DBs that you have access to. Excluding a database from dump-all can be a valid use case, but I'd go for something like `--exclude` instead.

Your confusion is exactly the confusion I'm taking about. `-d` isn't about including or excluding databases from `dump-all` it's connection option at the wrong place (it's only useful if you have dropped `edgedb` because otherwise, you there is no reason to specify `-d`, default would work).

text/1003-cli-naming.rst
ambv: There's still another suggestion @1st1 made during the meeting: keep them but make them different from \list-* in REPL: single-tab separated to be easily machine parseable with Unix pipelines.

ambv: (Keeping this as unresolved for now.)

text/1003-cli-naming.rst
ambv: Moving this to rejected ideas with Paul's rationale.

text/1003-cli-naming.rst
ambv: I'll be merging this PR with this section marked as "Under Consideration" as it looks like it needs more discussion.

Two comments:
1. Now I remember talks about "dump-all" and agree it should be a separate command. We can bikeshed the name though because IIUC "dump-all" is more like "dump-everything", including role passwords and port configuration, right? I would put it under `edgedb server dump` as this is where instance-level commands belong.
2. Yes, we have command-line option conflicts and that's ironically another reason why I'm +1 to this change. Why? Because those are confusing, especially when you're splitting a long command invocation in a Bash script or Dockerfile into multiple backslash-delimited lines. It's relatively easy to misread an option then or to edit a command and put an option in the wrong place. Removing all those conflicts would be ideal in my opinion.

text/1003-cli-naming.rst
ambv: Observation: not listing sub-commands alphabetically, per another accepted decision in the RFC, would help isolating the `list-*` sub-commands and so it might be enough in terms of `--help` readability. And we want to keep them to make machine-parseable versions of what you're getting in the REPL.

text/1003-cli-naming.rst
tailhook: > Now I remember talks about "dump-all" and agree it should be a separate command. We can bikeshed the name though because IIUC "dump-all" is more like "dump-everything", including role passwords and port configuration, right? I would put it under edgedb server dump as this is where instance-level commands belong.

I'm -1 on making it a `server` command. It works on any instance, not just local instances. Even more it's currently the only way for upgrading a remote instance.

We can do `dump-server` if `dump-all` is bad, but it may be confusing.

But dictionary says that "all" may be used as a synonym for "everything". So I think it short and clear enough.

1st1: > Yes, we have command-line option conflicts and that's ironically another reason why I'm +1 to this change. Why? Because those are confusing, especially when you're splitting a long command invocation in a Bash script or Dockerfile into multiple backslash-delimited lines. 

+1, good point.

1st1: > We can do dump-server if dump-all is bad, but it may be confusing.

It's `pg_dumpall` in Postgres, and IMO it would work just as good for us.

PR #28, opened 2021-02-16
Add subtransactions support

tailhook:
Major questions:
1. Name: `sub_transactions` or `savepoint`? (latter is bad for JavaScript at least)
2. Should `sub_transaction` be distinct from `try_transaction`?
3. Should transaction object be borrowed like connection is borrowed.


tailhook:
Another fun thing, is that PL/Python has [subtransaction](https://www.postgresql.org/docs/9.1/plpython-subtransaction.html) method/context manager.

elprans:
>     1. Name: `subtransactions` or `savepoint`? (latter is bad for JavaScript at least)

On this one I strongly prefer `subtransaction` since we are discussing a distinct _block_, as opposed to declaring an arbitrary _point_ to which you can rollback.

>     2. Should `subtransaction` be distinct from `try_transaction`?

The proliferation of different transaction block APIs worries me somewhat.  So, let me put it this way: what are the use cases for `try_transaction` and do they warrant a separate method?  Can we get away with just `retry()` and `subtransaction()` and say that if you want a non-retrying transaction, use `START TRANSACTION/COMMIT/ROLLBACK` explicitly (with a combination of `try/finally`)?

>     3. Should transaction object be borrowed like connection is borrowed?

Is there a reason to?  Also, I don't see clear rationale for borrowing on the main transaction, can we add that?

tailhook:
> The proliferation of different transaction block APIs worries me somewhat. So, let me put it this way: what are the use cases for `try_transaction` and do they warrant a separate method? Can we get away with just `retry()` and `subtransaction()` and say that if you want a non-retrying transaction, use `START TRANSACTION/COMMIT/ROLLBACK` explicitly (with a combination of `try/finally`)?

Explicit blocks are very dangerous in pools and reconnecting connections. The RFC states that we want to disable them for in `query()` too (along with disabling them in `execute()` which we have already done).

The reason for using `try_transaction` is still the same: if you need to implement your own retry strategy (this might happen for example if your transaction employs another storage and you can't split it into multiple chunks).

> Is there a reason to? Also, I don't see clear rationale for borrowing on the main transaction, can we add that?

I also don't see, and there is no borrowing described in PR. What should I add?

tailhook:
It looks like there are no objections here. Should we merge this?

1st1:
> It looks like there are no objections here. Should we merge this?

I'll also take a look tomorrow.

text/1004-transactions-api.rst
```diff
@@ -391,6 +391,53 @@ The ``RetryOptions`` signature:
...
+
+Subtransaction API is only accessible on the transaction object:
+
+.. code-block:: typescript
+
+    class Transaction {
+        async subtransaction<T>(action: TransactionBlock): Promise<T>;
+    }
+
+Example using ``subtransaction``:
```
1st1, line 405: For JS, maybe the name should be `subTransaction` (plays better with `tryTransaction`)?  Although Postgres calls it `subtransansaction` in their APIs and this is a single English word. @elprans ?

text/1004-transactions-api.rst
```diff
@@ -391,6 +391,53 @@ The ``RetryOptions`` signature:
...
+                    `, {name, text})
+                })
+            }
+            raise e
+        }
+    })
+
+Note:
+
+1. Transaction object is not "borrowed" for the duration of subtransaction,
```
1st1, line 432: What does this mean? Does it mean I can do `tx.subtransaction(tx => {` and while the closure is executing use `tx` in some other place?

1st1 (commented):
Overall looks great. Just a couple nit comments but pretty much good to go. Great stuff.

text/1004-transactions-api.rst
tailhook: Well, it doesn't make sense to use `tx` in "other place". But you use the same `tx` object for all queries and nested subtransactions. The note here is to show difference between subtransactions and `tryTransaction` / `retryingTransaction`.

tailhook: I'm still uncertain whether we want to pass the same object into subtransaction function in the JavaScript (or have zero-argument function there and rely on closure to propagate `tx` object).

1st1: > Well, it doesn't make sense to use tx in "other place"

Sure. But it doesn't mean one can't do that by mistake.

> I'm still uncertain whether we want to pass the same object into subtransaction function in the JavaScript (or have zero-argument function there and rely on closure to propagate tx object).

But why don't we create a separate "subtransaction object"? E.g.

```
for tx in con.retry():
   with tx:
      with tx.substransaction() as sub:
         sub.query()
```

?

tailhook: I don't have a good answer to that. I think there is less chance of loosing `tx` to another coroutine, just because it is not common to put transaction object to some global state. Also `tx` expires when transaction is closed. But yes, when the business logic is so large that it requires subtransactions it may have different constraints. Another argument for `transaction.query` is `pool.acquire`, and this argument doesn't apply to subtransactions too.

So this is mostly a matter if we want `subtransaction` to behave similarly.

tailhook: Well I almost forget our discussion about `subtx.rollback()`. So having that we probably need a real subtransaction object.

text/1004-transactions-api.rst
fmoor: two cents:

> plays better with tryTransaction

The motivation for this seems to be purely aesthetic and overlooks the function part of casing conventions. It doesn't take into account the camel casing pattern of making the first letter of words capital. This makes an arbitrary letter capital. It seems like this decision would make it harder for people to guess the correct spelling.

text/1004-transactions-api.rst
1st1: > So having that we probably need a real subtransaction object.

Yeah. @elprans Please read this discussion as it's API-related and confirm.

elprans: How would `subtransaction()` interact with `DECLARE SAVEPOINT/ROLLBACK TO SAVEPOINT`?  Will we ban them like we did with transaction statements?

As for subtransaction objects and enclosing transaction borrowing, I don't see why we shouldn't follow the same design as for transactions.

tailhook: > How would `subtransaction()` interact with `DECLARE SAVEPOINT/ROLLBACK TO SAVEPOINT`? Will we ban them like we did with transaction statements?

Yes. They are classified as transaction statements now (although banning them is still on to do list)

> As for subtransaction objects and enclosing transaction borrowing, I don't see why we shouldn't follow the same design as for transactions.

Okay.

tailhook:
Updated with "borrowing", rollback, and nested transaction example.

text/1004-transactions-api.rst
1st1: Good point, Frederick.

text/1004-transactions-api.rst
```diff
@@ -1057,6 +1227,40 @@ The implementation should reconnect and reapply the session
 configuration anyway. So it deviates from the usual behavior of context
 manager anyway.
 
+Seamless Subtransaction API
```
1st1, line 1230: It's not quite "seamless", IMO it's very explicit.

text/1004-transactions-api.rst
```diff
@@ -592,6 +685,83 @@ These base classes should be implemented by respective classes:
...
+                        UPDATE Note
+                        FILTER .name = <str>$name
+                        SET { text := <str>$text }
+                    ''', name=name, text=text)
+
+Note:
+
+1. Transaction object is "borrowed" for the duration of subtransaction,
+   you can't execute queries on original transaction object when subtransaction
+   is active.
```
1st1, line 728: Can we duplicate this block for the Connection API, clarifying that in

```
    async for tx in conn.retry():
        async with tx:
            try:
                async with tx.subtransaction() as subtx:
```

both `conn` qand `tx` cannot be used while `async with tx.subtransaction()` is running?

1st1 (approved):

text/1004-transactions-api.rst
tailhook: Added. Feel free to do post-merge edits if that is not clear enough.

PR #30, opened 2021-04-03
RFC 1005: CLI and conventions for local projects using EdgeDB

elprans:
This RFC describes the design of the ``edgedb project`` CLI commands for
the purposes of initialization of a software project that uses EdgeDB as
well as the associated conventions and changes to the language bindings.

### DIFF
--- a/text/1005-edgedb-project.rst
+++ b/text/1005-edgedb-project.rst
@@ -58,7 +58,7 @@ the ``edgedb project`` prefix:
 * ``edgedb project init`` -- populate a new project or initialize an existing
   cloned object;
 
-* ``edgedb project deinit`` -- remove association with and optionally destroy
+* ``edgedb project unlink`` -- remove association with and optionally destroy
   the linked EdgeDB instance;
 
 * ``edgedb project status`` -- show EdgeDB-related information about a
@@ -83,10 +83,13 @@ TOML::
 
     [edgedb]
     server-version = <semver-range>
+    schema-directory = <schema-dir>
 
-Here, the `server-version` attribute specifies a SemVer range of acceptable
+Here, the ``server-version`` attribute specifies a SemVer range of acceptable
 server versions.  Most frequently this would be in the form of a minimum
-required version.
+required version.  The ``schema-directory`` attribute specifies the location
+of EdgeDB schema directory relative to ``edgedb.toml``.  Defaults to
+``dbschema``.
 
 Project detection
 -----------------
@@ -120,8 +123,8 @@ The project -> instance mapping is expressed as directories under
 
 Here ``<dir-basename>`` is the trailing component of the path to the project
 directory, and ``<dir-abspath-hash>`` is defined as
-``lower(strip(base32(abspath(project_dir)), '==='))```.  Hashing is necessary
-to avoid bumping into the maximum directory entry name length.
+```lower(strip(base32(sha1(abspath(project_dir))), '==='))```.  Hashing is
+necessary to avoid bumping into the maximum directory entry name length.
 The ``project-path`` file contains the full absolute path to the project
 directory and is necessary for reverse lookup (find project by instance name),
 and the ``instance-name`` file contains the name of the associated instance.
@@ -143,7 +146,7 @@ Effect on ``edgedb server`` commands
 
 The ``edgedb server destroy`` command should refuse to destroy an instance that
 is associated with a project by default, and should recommend to use
-``edgedb project deinit`` (or ``edgedb server destroy --force``).
+``edgedb project unlink`` (or ``edgedb server destroy --force``).
 
 The ``edgedb server upgrade`` command should refuse to continue if the target
 server version does not match the ``edgedb.server-version`` range in
@@ -178,13 +181,15 @@ Options
   Specifies the EdgeDB server instance to be associated with the project.
   If the specified instance does not exist, it will be created.  If the
   specified instance already exists, it must not be associated with another
-  project.  ``edgedb server deinit`` may be used to disassociate an instance
+  project.  ``edgedb project unlink`` may be used to disassociate an instance
   prior to linking it with another project.
 
 ``--server-instance-type=<instance-type>``
-  Specifies the desired instance type.  Allowed values for ``<instance-type>``
-  are: ``native`` and ``docker``, which correspond to the installation modes
-  in ``edgedb server``.
+  Specifies the desired instance type.  Current allowed value for
+  ``<instance-type>`` is ``local`` (future additions include ``cloud``)
+
+``--server-install-mode=<install-mode>``
+  Corresponds to the local installation modes in ``edgedb server``.
 
 ``--non-interactive``
   Run in non-interactive mode.
@@ -278,7 +283,7 @@ Here's a simulation of a proposed interactive mode for a cloned project::
     Running migrations ...
 
 
-edgedb project deinit
+edgedb project unlink
 =====================
 
 Remove association with and optionally destroy the linked EdgeDB intstance.
@@ -286,7 +291,7 @@ Remove association with and optionally destroy the linked EdgeDB intstance.
 Synopsis
 --------
 
-``edgedb project deinit [options]``
+``edgedb project unlink [options]``
 
 Options
 -------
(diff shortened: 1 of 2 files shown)


1st1 (commented):
Good job!

Just a few comments:

1. Please describe in detail what `edgedb project status` will output.

2. Why `lower(strip(base32(abspath(project_dir)), '==='))` and not `sha1(abspath(project_dir))`? The latter would produce shorter names and the dirname would still be distinguishable because of the basename prefix.

3. I don't 100% like `project deinit` command name because it's not a direct opposite of `project init`. I.e. `init` creates `edgedb.toml` for you, I'd expect `deinit` to remove it. Maybe we call it `project unregister`?  Although this is a minor issue and I'm OK if we end up calling it `deinit`.

4. In the "Use a globally-unique identifier for projects" section I'd mention that we don't actually need a globally unique ID, we need a locally unique one for the given devbox. Our `edgedb.toml` hashing algorithm produces exactly that.

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+----------------------
+
+`RFC 1000 <1000-migrations.rst>`_ defined the top-level ``dbschema/`` directory
+as the default location of EdgeDB SDL schema, and ``dbschema/migrations`` as
+the default expected location of the files containing schema migration scripts.
+To mark a directory as an EdgeDB project, an ``edgedb.toml`` file should be
+present in the root directory.  The file can be empty or contain the following
+TOML::
+
+    [edgedb]
```
tailhook, line 84: What does the section means? Is this sorta comment? Does it mean we should put everything under `edgedb` in the future, i.e. `[edgedb.extensions]`, etc.

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+``~/.edgedb/projects``::
+
+    ~/.edgedb/projects/
+        <dir-basename>-<dir-abspath-hash>/
+            project-path
+            instance-name
+
+Here ``<dir-basename>`` is the trailing component of the path to the project
+directory, and ``<dir-abspath-hash>`` is defined as
+``lower(strip(base32(abspath(project_dir)), '==='))```.  Hashing is necessary
```
tailhook, line 123: Looks like a typo. There is no hasing?

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+The ``edgedb project init`` command initializes a brand new project or
+re-initializes an existing project.
+
+In a new project:
+
+- an ``edgedb.toml`` file is created in the project directory,
+  and ``--server-version``, if specified, is recorded in the
+  ``edgedb.server-version`` attribute.
+
+- a ``dbschema`` directory and a ``dbschema/default.esdl`` file are created,
```
tailhook, line 204: How "new project" is detected? Is it edgedb.toml? It's unspecified what we do if there is no `edgedb.toml`, but there are something in `dbschema`

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+  the latest nightly version.
+
+``--server-instance=<instance>``
+  Specifies the EdgeDB server instance to be associated with the project.
+  If the specified instance does not exist, it will be created.  If the
+  specified instance already exists, it must not be associated with another
+  project.  ``edgedb server deinit`` may be used to disassociate an instance
+  prior to linking it with another project.
+
+``--server-instance-type=<instance-type>``
```
tailhook, line 184: Why `instance-type` and not `install-method` like we use elsewhere? Should we rename `server install --type` ?

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+
+Options
+-------
+
+``--project-dir=<dir>``
+  Specifies a project root directory explicitly.  If not specified, the project
+  directory is detected as described in the "Project detection" section above.
+
+``--destroy-server-instance, -D``
+  If specified, the associated EdgeDB instance is destroyed by running
```
tailhook, line 299: What is the purpose of deinit without destroy? I guess it will just leave danging instances where users expect them to be removed.

text/1005-edgedb-project.rst
```diff
@@ -0,0 +1,418 @@
...
+
+
+Implementation
+--------------
+
+The ``edgedb project status`` shows the following information about the
+project:
+
+- associated EdgeDB instance name, or `<none>` if not associated;
+- migration status (which deprecates ``edgedb show-status``)
```
tailhook, line 342: Why would we deprecate `show-status`? What to do with remote/cloud instances?

tailhook (commented):
The most important is that I don't understand version management:

> Defaults to '*', which means latest stable version for new projects

>  an edgedb.toml file is created in the project directory, and --server-version, if specified, is recorded in the edgedb.server-version attribute.
    
So what is written to `toml`? Is it `*`? Or is it empty file?
    
The second thing is that *we aren't semver-compatible*: `1.0b1-2021030102.el7` < `1.0b1` as of semver. Also libraries managing semver usually require three components. I don't see it required in the spec, but it's somehow implied in examples. So basically we can't use semver tools.

elprans:
> So what is written to `toml`? Is it `*`? Or is it empty file?

`server-version = "*"` is written by default, meaning that any stable EdgeDB version is acceptable, and if none is installed, the latest would be.  As for how to treat an empty file: we can either make that an error, or treat it as equivalent to the above.

> The second thing is that _we aren't semver-compatible_

Good point, although that is only partly true.  The internal server version _is_ SemVer, what you are looking at are .deb and .rpm versions that follow different rules.  We need to add a "semver" field to JSON indexes.

text/1005-edgedb-project.rst
elprans: We could call this `[server]` for clarity.

> Does it mean we should put everything under `edgedb` in the future, i.e. `[edgedb.extensions]`, etc.

If it's server-related, sure.  We can potentially have other sections that drive the CLI or other clients.

text/1005-edgedb-project.rst
elprans: Yes, a typo, I meant to use `base32`, will fix.

text/1005-edgedb-project.rst
elprans: > there is no `edgedb.toml`, but there are something in `dbschema`

Presence of `dbschema` no longer indicates a "project", only `edgedb.toml`.  If `dbschema` already exists, we simply don't attempt to create it.  I'll clarify.

text/1005-edgedb-project.rst
elprans: This is a bit different.  `install-method` refers to the local installation method (native vs docker), whereas instance type specifies `local` vs `cloud`, although in this case `local` branches to include the installation methods.  We can make a separate `--install-method` option instead and keep `--instance-type` as `{'local`, 'cloud'}`.

text/1005-edgedb-project.rst
elprans: It is mainly to fix unintentional linking.  If you picked a wrong instance by mistake, this will allow you to unlink it.

text/1005-edgedb-project.rst
elprans: `show-status` is about migrations, you are thinking about `server status` here.  That said, there is still value in the ability to query the status of migrations on an instance without an explicit project.  That said, we should still deprecate `show-status` as we've agreed to rename it to `migration-status` for clarity (see RFC 1003).

text/1005-edgedb-project.rst
1st1: Would also be cool to be able to configure a different dir for `dbschema` in `edgedb.toml`. For projects that use multiple databases simultaneously, `dbschema` might already be taken or just be confusing.

text/1005-edgedb-project.rst
elprans: I've split the option into two.

text/1005-edgedb-project.rst
elprans: Added

1st1 (approved):

PR #33, opened 2021-05-19
Initial draft of RFC 1006

1st1:


text/1006-groupped-cli.rst
```diff
@@ -0,0 +1,240 @@
...
+    The `create-migration` command has been deprecated.
+    Use `edgedb migration create` instead.
+
+
+Open Questions
+==============
+
+* Do we need to remove or redesign ``edgedb configure``?
+
+* Do we need ``edgedb query``?
```
tailhook, line 235: Sure. What's wrong with `edgedb query`?

text/1006-groupped-cli.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    dump                       Create a database backup
+    restore                    Restore a database backup from file
+    configure                  Configure a DB or a server instance
+
+    migration apply            Migrate the database to the latest revision
+    migration create           Create a new migration
+    migration log              Show the migrations log
+    migration status           Show the current migration state
+    migrate                    An alias for `edgedb migration apply`
```
tailhook, line 91: This "convenient" alias will make completion uglier (you always have to type `m<tab>i<tab>`

text/1006-groupped-cli.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    project init               Initialize a new EdgeDB project
+    project status             Show the status of the current project
+    project list               List all projects
+    project unlink             Clean-up the project configuration
+
+    server status              Status of an instance
```
fantix, line 98: ### Thoughts

`edgedb server status` without instance name feels more like `edgedb server list-instances` intuitively, especially when we also have `edgedb server info` (and `edgedb server list-versions`) which sounds similar but actually shows info about the installations. Even though I know it's an over-optimization to suggest a new `edgedb instance` group for all instance-related subcommands (because instances depend on installations, and they are all "servers" literally), but `edgedb server status` keeps pushing me towards that direction.

### Proposal

Perhaps `edgedb server list-instances` could be an alias of {`edgedb server status` without instance name}?

fantix (commented):
As a partial tech-neat-freak I even want to over-optimize the `edgedb server` group into `edgedb instance`, so I'm +1 for this RFC - it does make me feel relaxed.

text/1006-groupped-cli.rst
tailhook: The initial plan was to have `server status --all`. But it turned out that `status` without name is useless and could print a list instead of erroring out. Also we expect most of the users having a small number of instances (probably single one). So it doesn't make sense to learn two commands.

We may have `list-instances` that is faster as does no status check for scripting.

text/1006-groupped-cli.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    $ edgedb create-migration
+    The `create-migration` command has been deprecated.
+    Use `edgedb migration create` instead.
+
+
+Open Questions
+==============
+
+* Do we need to remove or redesign ``edgedb configure``?
```
1st1, line 233: Will rename this to `edgedb config`.

text/1006-groupped-cli.rst
```diff
@@ -0,0 +1,240 @@
...
+proposes. There has to be a group for all ``list-X`` and ``describe-Y``
+commands and ``edgedb inspect`` seems like a reasonable choice.
+
+It is important understand the context of how the inspect commands are used.
+Most of the time they are being accessed from REPL, e.g.::
+
+    db> \lt                     # list object types shorcut
+    db> \list-object-types      # the full command
+
+Using them from the CLI is less likely, but not exposing the to the CLI at
```
vpetrovykh, line 153: > but not exposing the to the CLI

"the" -> "them"?

text/1006-groupped-cli.rst
jaclarke: > (Is this already implemented? I don't have autocompletion on my machine.)

You may need to run the gen_completions command first: `edgedb _gen_completions --help`

text/1006-groupped-cli.rst
vpetrovykh: While we're on the subject of tweaking the CLI `configure` group, maybe we should also rename `configure insert` to `config add`.

The reasons for the change:
* `insert` sounds like a DB command that often goes together with `delete` and `update`. At least in my experience it did make me look for those other commands in vain and being unsure whether the docs are outdated, or whether the implementation is incomplete. We can still leave the commands in EdgeQL as they are, but at least for the CLI, we might as well avoid the confusion.
* We already are breaking from an exact match between the CLI config commands and EdgeQL `CONFIGURE` commands and we intend to do more of it, specializing `auth` commands into their own CLI group, etc.
* I would argue that we should look at the situation from a perspective of a new user, who isn't aware of all the intricacies of EdgeQL yet. Thus they probably aren't even aware of the `CONFIGURE` commands, they are just trying to figure out how they can change some configuration aspect and `edgedb config --help` is possibly their first stop on this path. Seeing `insert`, `set` and `reset` may be a bit more confusing than `add`, `set`, `reset` when the intent is to either "add" some new config or "update" an existing one. "Updating" or "deleting" configuration is especially interesting in that in the presence of an `insert` command, `set` and `reset` look less like viable candidates for those actions, whereas if the command is a non-keyword `add`, then there's no built-in expectation to match it with other keywords, so `set` and `reset` seem more natural.

I realize that this is very subjective, but I think it's worth considering since even members of our own team got a little confused by the `configure insert`.

text/1006-groupped-cli.rst
1st1: fixed

text/1006-groupped-cli.rst
tailhook: They should be generated on installation. What's your shell?

tailhook: > worth introducing a highly visible breaking change 

We are breaking almost every other command in a very visible way. So keeping just this one for the sake of compatibility sounds like a far-fetched argument. Other than that, I don't know if it worth it.

text/1006-groupped-cli.rst
tailhook: Well, `-c` used in command which don't have subcommands. I.e. `sh -c` is there because `sh exec` would search for a file named `exec`. Same with `python`, etc.

But we have subcommands, so some subcommand makes more sense. Also the subcommand can have additional nice arguments like `edgedb query -f file.esdl` or `edgedb query --format=json` or `edgedb query --show-times`, which would be ugly to have as global arguments (well, we already sorta have format, but I'm not sure this is a great idea).

colinhacks:
Proposal w.r.t. `inspect`. The `edgedb inspect list-X` feels a little redundant, I think this would be easier to learn and more understandable:

```
edgedb inspect [TARGET]

TARGETS:

  schema    Show the schema in SDL
  aliases       List type aliases
  casts         List casts
  databases     List databases
  indexes       List indexes
  modules       List modules
  roles         List roles
  object-types  List object types
  scalar-types  List scalar types

  -t <type>           Describe the matching DB object
```

Thoughts?

tailhook:
@colinhacks, the problem with that is slash commands. Currently we have `\la, \list-aliases`. I.e. we assume that `\inspect` is stripped for the REPL commands. In your proposal it's not clear what they should be:

*  `\i a, \inspect aliases` -- well, space does make `\i a` weird
* `\ia, \inspect aliases` -- makes `\ia` work differently, not as a subcommand. Also `\inspect aliases` reads as `\inspect <space-separated-list-of-targets>` instead of literally
* `\la, \inspect aliases` -- hard to remember and same issue as above

PR #36, opened 2021-06-16
DRAFT: add TLS and ALPN support

fantix:
This pre-RFC is for early discussion and not a proper RFC yet, because it contains ambiguity and undecided options and questions, and is not yet following RFC 0001 properly.

[rendered](https://github.com/fantix/rfcs/blob/tls/text/tls-and-alpn.rst)

UPDATE: Thank you all for the valuable comments! I'll summarize the outstanding questions or current decisions below for the next draft, and I'll close this PR as it is and start a new one on a new branch, so that the discussion history can be preserved.

1. Use a local CA or issue self-signed certificates separately for each server instance? CA is easier to trust a single root certificate, and separate self-signed certificates is just cleaner like how SSH handles the server fingerprints.
2. Store optional TLS settings in server config, don't use command parameters.
3. UNIX socket: remove non-admin ones, clear text for admin socket and use the binary protocol only.
4. We probably don't need passphrasses on the generated private keys?
5. The client should transparently work with dev certs and prod certs, no additional switch is to be added.
6. If an ALPN negotiation didn't succeed, error out.
7. Optional TLS settings should have the same default as Python does.
8. If the user brings cert files with them, we should just use them in place. Importing/copying adds more unnecessary work to us.
9. NO for integrating with letsencrypt or something similar.
10. Supporting client certificate authentication is a nice to have for this RFC.
11. Who eventually generates the certificates? The Python server or the Rust CLI? Sounds like we'll use the Rust CLI.
12. Which ALPN protocol option? Option 1.


text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+SSLv2 and SSLv3 are not supported. TLSv1.3 support is included (question:
+shall/can we deliver with OpenSSL >= 1.1.1 consistently for all platforms?).
+
+``--tls-certfile``, ``--tls-keyfile`` and ``--tls-private-key-password`` are
+directly the parameters of Python ``ssl.SSLContext.load_cert_chain()``,
+except that the password cannot be a function. Instead, the EdgeDB server would
+accept password as an environment variable ``EDGEDB_TLS_PRIVATE_KEY_PASSWORD``.
+
+If specified, ``--tls-ciphers`` will be used to call
+``ssl.SSLContext.set_ciphers()``.
```
1st1, line 103: Also, do we need this as a parameter? IMO it makes sense to make this a config entry (that's what Postgres does with `ssl_ciphers` https://www.postgresql.org/docs/10/ssl-tcp.html)

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
+TLS and ALPN
+============
+
+This RFC proposes to change the transport of EdgeDB frontend connections to TLS,
+and use ALPN for multiplexing protocol selection.
+
+
+Motivation
```
1st1, line 8: BTW, we have UNIX sockets (and AFAIK we'll continue to have them) -- IMO we should reserve UNIX sockets for clear-text binary proto. cc @elprans

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+reliable protocol selection on top of the TLS transport, allowing EdgeDB to
+multiplex different frontend protocols like the binary protocol, EdgeQL over
+HTTP or GraphQL over HTTP (and maybe more in the futre) on the same port as
+recommended in Pre-RFC 1910 [3]_.
+
+
+Overview
+--------
+
+EdgeDB server CLI (``edb server``) will take a few more parameters to configure
```
1st1, line 33:
```suggestion
EdgeDB server CLI (``edgedb server``) will take a few more parameters to configure
```

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+If a specific protocol was successfully negotiated, that protocol will be used
+as the exclusive one throughout the lifetime of the connection. Or else if the
+ALPN agreement was not made, depending on the discussion below we will fallback
+to either the EdgeDB binary protocol or the current HTTP protocol.
+
+
+Development of EdgeDB
+---------------------
+
+As mentioned in the Overview, the ``edb server`` CLI will take extra parameters
```
1st1, line 82:
```suggestion
Just like ``edgedb server``, the ``edb server`` CLI will take extra parameters
```

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+certificate is usually sufficient, so the CLI tooling should have an option to
+automatically generate that for the user transparently. For production, we'd
+also like to make the certificate/key management easier and safe, options are
+discussed later.
+
+
+Internals
+^^^^^^^^^
+
+Under the neath, we'll have new files and folders in the EdgeDB well-known
```
1st1, line 129:
```suggestion
We'll have new files and directories in the EdgeDB well-known location
```

?

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+        "database": "edgedb",
+        "tls_private_key_password": "...."
+    }
+
+Question: I don't feel comfortable storing the key passphrase here in clear
+text, especially if this is in production environment and we asked the user
+for passphrase but stored in this file. Perhaps we could use this JSON value
+only for the development process, and introduce a more secure (thus less
+convenient) way to obtain the private key password, like a customizable command
+as in PostgreSQL?
```
1st1, line 174: Yeah, -1 to storing passwords in clear text for any purposes. We can add an env var for that which is a bit better,

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+in production we'll "import" the certs specified by the user, by copying and
+renaming the cert and key files to the ``certs/`` folder.
+
+
+Development Scenario
+^^^^^^^^^^^^^^^^^^^^
+
+From the user's perspective, ``edgedb server init`` (without parameters we are
+introducing in this RFC later) will now automatically generate a self-signed
+certificate with a random key passphrase, and set the above config properly.
```
1st1, line 188: Perhaps we should ask the user to enter the passphrase (or supply it via a command line arg or an env var)

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+``$WELLKNOWN/credentials/<instance>.json``. We could update our client bindings
+and the CLI to honor this flag, load and trust the certificate so that the user
+doesn't have to write extra code in development. Though, this works only for
+local clients - remote clients don't have access to the local file system so
+there has to be a different way for them to trust the dev server. If this
+option is good, I will expand ths paragraph with details. Alternatives:
+
+1. Automatically build a local CA, and issue certificates from that CA. The
+   clients and CLI will only need to trust the CA certificate. This doesn't
+   solve the remote client trouble though.
```
1st1, line 207: Yeah, I think this should be the default way to go for both `edb server` and `edgedb server` workflows.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+2. Provide a client-side switch that the user could easily set, so that the
+   user could explicitly turn it on for development mode, and off for
+   production.
```
1st1, line 210: I don't think we need a switch -- just good documentation explaining how to deploy in docker with user-provided certs.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+
+Overview
+--------
+
+EdgeDB server CLI (``edb server``) will take a few more parameters to configure
+a proper TLS server::
+
+    --tls-certfile
+    --tls-keyfile
+    --tls-private-key-password
```
1st1, line 38: certfile, keyfile, and passwd should also be configurable via env vars.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+preferred protocol for each connection. The ALPN support in target languages:
+
+* Python [4]_: ``set_alpn_protocols()`` and ``selected_alpn_protocol()``
+* Go [5]_: ``SupportedProtos`` and ``NegotiatedProtocol``
+* Node.js [6]_: ``ALPNProtocols`` and ``alpnProtocol``
+
+If a specific protocol was successfully negotiated, that protocol will be used
+as the exclusive one throughout the lifetime of the connection. Or else if the
+ALPN agreement was not made, depending on the discussion below we will fallback
+to either the EdgeDB binary protocol or the current HTTP protocol.
```
tailhook, line 76: Don't see why we wouldn't make ALPN required.

We may try to detect clear-text connections for some cases. But maybe even this is not needed.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+As mentioned in the Overview, the ``edb server`` CLI will take extra parameters
+for TLS. We will use the Python builtin TLS support to handle the certificates
+and ALPN, and use the TLS transport implementation in uvloop for the network.
+
+The server-side ``ssl.SSLContext`` [4]_ will be initialized with the default
+``protocol=ssl.PROTOCOL_TLS``, leaving the control of accepted TLS protocol
+versions to ``SSLContext.minimum_version`` and ``SSLContext.maximum_version``,
+which in turn are managed by:
+
+* ``--tls-protocol-min``: One of ``1.0`` (default), ``1.1`` and ``1.2``
```
tailhook, line 91: Why 1.0 is the default? Are any language bindings that don't support 1.2?

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+* ``--tls-protocol-max``: One of ``1.0``, ``1.1``, ``1.2`` (default) and ``1.3``
```
tailhook, line 92: What's wrong with 1.3? I think default should whatever is supported by our implementation.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+recommended in Pre-RFC 1910 [3]_.
+
+
+Overview
+--------
+
+EdgeDB server CLI (``edb server``) will take a few more parameters to configure
+a proper TLS server::
+
+    --tls-certfile
```
1st1, line 36: Perhaps if no `--tls-certfile` was provided we always auto-generate a local one?

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+
+Question: do we want to manage the certificate for the user? That means to
+import the user certificate files and store locally, warn about expiry or even
+run certbot to renew some free certs (this might be too much). If yes, the
+following in this section is just one of the options.
+
+``edgedb server init`` will have additional parameters to import the TLS
+certificate and key files for the new EdgeDB instance::
+
+    --import-tls-certfile
```
1st1, line 224: Maybe we should always import if `--tls-certificate` is supplied with a user-provided cert?

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+---------------------
+
+As mentioned in the Overview, the ``edb server`` CLI will take extra parameters
+for TLS. We will use the Python builtin TLS support to handle the certificates
+and ALPN, and use the TLS transport implementation in uvloop for the network.
+
+The server-side ``ssl.SSLContext`` [4]_ will be initialized with the default
+``protocol=ssl.PROTOCOL_TLS``, leaving the control of accepted TLS protocol
+versions to ``SSLContext.minimum_version`` and ``SSLContext.maximum_version``,
+which in turn are managed by:
```
tailhook, line 89: Do we want to put some of them into server config? Certainly not the path to the keyfile and password. But min/max protocol and ciphers could be easily changed at runtime and backed up in a regular database backup.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+
+Option 2: Fine-grained Protocols
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+All protocols/extensions are given a unique ALPN protocol name, so that the URI
+can be largely simplified. Other than that, the actual protocols remain
+unchanged. The server will advertise all ALPN protocols for simplicity, even
+though some extension may never be configured on any database.
+
+If for some reason the ALPN negotiation didn't work, the server will fallback
```
1st1, line 279: What can be a reason for ALPN to not work?

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+don't have to maintain 2 sets of URI structures.
+
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| Protocol Name     | Description                | Example URI                        | Equivalent URI in old server       |
++===================+============================+====================================+====================================+
+| ``edgedb-binary`` | The EdgeDB binary protocol | N/A                                | N/A                                |
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| ``system-api``    | Server system API          | ``/status/ready``                  | ``/server/status/ready``           |
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| ``edgeql+http``   | EdgeQL over HTTP           | ``/<database>?query=``             | ``/db/<database>/edgeql?query=``   |
```
1st1, line 291:
```suggestion
| ``edgeql+http``   | EdgeQL over HTTP           | ``/<database>/edgeql``             | ``/db/<database>/edgeql``   |
```

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| ``graphql``       | GraphQL over HTTP          | ``/<database>/explore``            | ``/db/<database>/graphql/explore`` |
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| ``notebook``      | The Notebook extension     | ``/<database>/status``             | ``/db/<database>/notebook/status`` |
++-------------------+----------------------------+------------------------------------+------------------------------------+
+| <default>         | No ALPN protocol match     | ``/server/status/ready``           |                                    |
+|                   |                            |                                    |                                    |
+|                   |                            | ``/db/<database>/edgeql?query=``   |                                    |
```
1st1, line 299: ditto

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+        |    +--<instance>.json       <UPDATED>
+        |--certs/                     <NEW>
+        |    +--<instance>.cert.pem   <NEW>
+        |    +--<instance>.key.pem    <NEW>
+        ...
+
+The ``config/tls.json`` contains optional TLS settings for each instance::
+
+    {
+        "<instance>": {
```
tailhook, line 146: I.e. this part can be avoided if put into a server config per comment above.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+---------------------------
+
+We mentioned that the new client is recommended to have an option (with
+warnings) to connect to old EdgeDB servers without TLS. In the other hand, we
+won't allow old EdgeDB clients connecting to newer EdgeDB servers with TLS
+enabled. On a TLS handshake failure, the EdgeDB server will return a error in
+clear text, indicating that the client must be upgraded to use TLS.
+
+However, the core EdgeDB server (``edb server``) will provide a hidden option
+``--no-tls`` to run the server in non-TLS compatible mode for development and
```
1st1, line 342: or perhaps we just use unix sockets for local dev...

text/tls-and-alpn.rst
tailhook, line 174: It's not more valuable then database password, so probably ok.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+for passphrase but stored in this file. Perhaps we could use this JSON value
+only for the development process, and introduce a more secure (thus less
+convenient) way to obtain the private key password, like a customizable command
+as in PostgreSQL?
+
+Under the new folder ``certs/``, we'll place all certificates and keys used by
+the EdgeDB instances. The cert files will be named after the instance names.
+For development scenario we'll likely generate those certs for the user, while
+in production we'll "import" the certs specified by the user, by copying and
+renaming the cert and key files to the ``certs/`` folder.
```
tailhook, line 180: Does named cert file includes whole chain? Also this assumes that client certificates are used, right?

tailhook (commented):
Few comments on the first part. I have to go now, will continue later.

text/tls-and-alpn.rst
1st1: > Don't see why we wouldn't make ALPN required.

+1.

text/tls-and-alpn.rst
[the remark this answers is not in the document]
fantix: > Why do we need these switches? 

I just picked two ("TLS protocol versoin range" and "available ciphers") settings that I thought are important enough to be customizable. Like you said in the next comment, we could just use a config entry. They don't have to be CLI parameters.

Postgres put them in the [runtime config](https://www.postgresql.org/docs/13/runtime-config-connection.html#RUNTIME-CONFIG-CONNECTION-SSL):

> ssl_min_protocol_version (enum)
>
> > Sets the minimum SSL/TLS protocol version to use. Valid values are currently: TLSv1, TLSv1.1, TLSv1.2, TLSv1.3. Older versions of the OpenSSL library do not support all values; an error will be raised if an unsupported setting is chosen. Protocol versions before TLS 1.0, namely SSL version 2 and 3, are always disabled.
> > 
> > The default is TLSv1.2, which satisfies industry best practices as of this writing.
> 
> ssl_max_protocol_version (enum)
>
> > Sets the maximum SSL/TLS protocol version to use. Valid values are as for ssl_min_protocol_version, with addition of an empty string, which allows any protocol version. The default is to allow any version. Setting the maximum protocol version is mainly useful for testing or if some component has issues working with a newer protocol.

text/tls-and-alpn.rst
fantix: For the record, in a video discussion we agreed to drop the UNIX sockets for non-admin purposes, and keep the admin UNIX socket with clear-text binary protocol only. Meanwhile, we'll drop the (hopefully already deprecated) admin connection option in all bindings, and the CLI will be the only client that connects to the admin UNIX socket.

text/tls-and-alpn.rst
fantix: For the self-signed certificate? Sure we can have the 3 ways to collect a passphrase, but I'm not quite getting the idea to have passphrase for development 🤔

text/tls-and-alpn.rst
fantix: Works for me 👍 And for the implementation, I think both could use a new command like `edb ca` that uses OpenSSL (new dependency) to operate the CA and issue dev certs. (No more rustls?)

text/tls-and-alpn.rst
fantix: Yes that's the idea. I'll clarify it.

text/tls-and-alpn.rst
fantix: You mean even for `edb server` we do the import? Let me update the RFC with the conclusions and suggestions for now and let's continue this discussion there.

text/tls-and-alpn.rst
fantix: Oh sorry - this is an example where a 3rd-party client didn't specify any ALPN protocol and the negoation was a failure.

text/tls-and-alpn.rst
fantix: Sorry about the ambiguous wording! I think it also works for me to raise an error if a 3rd-party client fails to provide a valid ALPN protocol, if that's what you meant please?

text/tls-and-alpn.rst
fantix: > Why 1.0 is the default?

Oh this is because I think `ssl.SSLContext.minimum_version` has a default of `ssl.TLSVersion.MINIMUM_SUPPORTED` and that is effectively TLS 1.0. I'd like to be as compatibile as Python is, though I think Postgres made 1.2 as the default minimum version, so I'm actually okay with either.

text/tls-and-alpn.rst
fantix: > What's wrong with 1.3?

This scared me: https://docs.python.org/3/library/ssl.html#tls-1-3:
>     Python has provisional and experimental support for TLS 1.3 with OpenSSL 1.1.1.


> I think default should whatever is supported by our implementation.

Yeah I agree - I'll make the default consistent as Python's `ssl.TLSVersion.MAXIMUM_SUPPORTED`.

text/tls-and-alpn.rst
fantix: > Does named cert file includes whole chain?

No I wasn't planning to include the root CA certificate. But the user should be responsible for imporing certfile with the necessary intermediate certificate in the chain for their client certificate verification to pass. So in other words, our client bindings trust only the dev CA root certificate (for dev only) in addition to the system CA repo, and the certificate verification on the client side depends on what's in the named cert file.

> Also this assumes that client certificates are used, right?

Sorry about the confusion! I didn't mean to involve client certificates. That's a good reminder though that we should also think about managing trusted client certificates. I'll add a new section for that, and the implementation may have a lower priority I guess.

text/tls-and-alpn.rst
tailhook: Yes.

text/tls-and-alpn.rst
tailhook: 1. We don't need this password if we don't have client certificates. Which @fantix said we don't want in another thread.
2. But we need database password. Actually credentials file was created specifically to store secrets and must be saved on the secrets partition, so it doesn't make much sense to move password out of this file. On the contrary, everything other could be moved out from this file to a `EDGEDB_DSN` or dsn specified by other means.

text/tls-and-alpn.rst
tailhook: Oh, sorry I've misread this. So:

> in production we'll "import" the certs specified by the user, by copying and renaming the cert and key files to the ``certs/`` folder.

I don't think this makes sense. If user specified path to file on the command-line they assume that file is used, not copied somewhere. And if it copied, then when? (On each startup? Why whould you need copy in this case). Generally, let users manage private files, they will probably use their own private partitions for that.

tailhook: Also: client needs some way to discover root CA certs of the servers chain. This must either be a well-known location, or `credentials.json` itself.

Unless, we rely on universal root CAs on production, and use lets encrypt. Which this spec doesn't say we use. (And it's bad for some well-firewalled on-premise installations).

text/tls-and-alpn.rst
tailhook: I agree, we don't need a passphrase here.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+From the user's perspective, ``edgedb server init`` (without parameters we are
+introducing in this RFC later) will now automatically generate a self-signed
+certificate with a random key passphrase, and set the above config properly.
+This is considered to be the development scenario, so we try not to let the
+user do extra work here - the ``edgedb server init`` command shouldn't ask
+questions about generating the self-signed certificate. Instead, it should warn
+about the self-signed certificate, and provide suggestions for production.
+
+Question: because the certificate is self-signed, certificate verification on
+the client side would fail. Now that we know we are "in development mode"
```
tailhook, line 195: > Now that we know we are "in development mode"

I don't think we know. Do we consider localhost a development mode? Docker installation is not always localhost.

text/tls-and-alpn.rst
tailhook: Rustls has CA functionality. But do you mean `edgedb ca` ? (`edb` is a developement command, I don't think you propose to bundle some CA key in to the distribution itself, right?)

I don't think we need CA. Just put self-signed certificate (public part) into a `~/.config/edgedb` folder somewhere and use that in-client.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+   user could explicitly turn it on for development mode, and off for
+   production.
+
+
+Production Scenario
+^^^^^^^^^^^^^^^^^^^
+
+Question: do we want to manage the certificate for the user? That means to
+import the user certificate files and store locally, warn about expiry or even
+run certbot to renew some free certs (this might be too much). If yes, the
```
tailhook, line 218: If by `certbot` you mean `letsencrypt` (its probably possible to use certbot for other compatible services), then I don't think it will work for more that 20% of use case. There are few ways certbot can confirm it's host to get certificate, most commonly:
1. By exporting some files over HTTP
2. By modifying DNS entries.

(1) is a problem because of two things: (a) users will hesitate exposing HTTP to the whole world on their database server and (b) database servers may not have externally available DNS names or IP addresses at all (e.g. have some `.local` name and `10.x.x.x` address).

(2) is a problem because of access control for DNS is usually not that fine-grained, and giving your database server access to the whole domain is not a great idea.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+following in this section is just one of the options.
+
+``edgedb server init`` will have additional parameters to import the TLS
```
tailhook, line 221: `server init` is not for production (at least yet). So I don't see why you would ever want to import certificates here. Just always use generated ones (until we have `server init --system` or something, if ever).

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+    --tls-private-key-password
+
+Again the password option here yields a security concern as mentioned
+previously. We should probably follow the way PostgreSQL used to make it a
+customizable command to obtain the passphrase.
+
+For simplicity, optional TLS settings are not provided on ``server init``.
+These options will also be added to ``edgedb project init`` command.
+
+In addition, we should provide a new ``edgedb server update-tls`` command for
```
tailhook, line 235: If we want updating cli params, we should have non-TLS specific command (`edgedb server update-config` or something), and implement those parameters in it.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+protocol in the TLS handshake, and be ready to handle any negotiation result,
+depending on the conclusion we'll make for previous questions.
+
+For compatibility considerations, the client is recommended to provide a
+temporary option to connect to old EdgeDB servers without TLS. However, this
+option should come with a warning and should not be default.
+
+For certificate verification, this was discussed previously. To recap, in order
+to make the client code simple and consistent, the client should either check
+the well-known EdgeDB location for the flag of self-signed certificate in order
```
tailhook, line 326: Let's have a well-known location of self-signed certificate itself, not to the flag, as described above. Then client don't have to hardcode a switch between verified and non-verified mode. It's always verified.

Specifically for CLI we can ask if we don't have a certificate yet, similarly to how SSH behaves on the first connect. But probably it's better to have certificate (or checksum of thereof) in the `credentials.json`. So that only file is enough to connect.

text/tls-and-alpn.rst
tailhook: Let's just make `edb server` initialize self-signed certificate and put credentials file like `local-dev.json` (or whatever). So you can use `edgedb -Ilocal-dev` to connect to `edb server`.

text/tls-and-alpn.rst
```diff
@@ -0,0 +1,380 @@
...
+    Generate (and update) self-signed certificate for development and testing.
+
+        -D, --data-dir  path to the EdgeDB data directory where the generated
+                        certificate and key files will be stored
+
+But this doesn't work for ``edb server --temp-dir`` or temporary servers
+started in the test suite. Therefore the new proposal is to just generate new
+self-signed certificate while the server is starting in devmode or testmode if
+no exising certificate is found or given. For that, we'd need either the
+OpenSSL library as a new development dependency, or do it through a Rust
```
tailhook, line 368: I think we're going to embed cert generation in CLI anyway. So let's just use the CLI.

Codewise cert generation is not that big, we already do that in tests: https://github.com/edgedb/edgedb-cli/blob/master/tests/certs.rs (I'm not sure if binary size of that is big, but I don't think so).

tailhook (commented):
Done first pass.

I would also like to have client certificates supported eventually, but I'm okay we don't do this right now.

text/tls-and-alpn.rst
fantix: > But do you mean `edgedb ca` ?

Oh sorry I meant the general Python code, I was confused. You're right, we could use the CLI to operate the CA - not to say I think we need one.

PR #37, opened 2021-06-18
RFC 1008: TLS and ALPN Support

fantix:
[rendered](https://github.com/fantix/rfcs/blob/tls-v2/text/1008-tls-and-alpn.rst)

Closes #36


text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+
+And echo back the generated certificate in standard output so that the
+server could simply read and use. So that the EdgeDB developer could
+always use ``edgedb -Ilocal-dev`` to access the dev server. The test
+suite could also take advantage from this mimic of real-life EdgeDB
+scenario to cover some real cases.
+
+Another server-side topic that was discussed in this RFC is the UNIX
+domain socket. It is proposed that the non-admin UNIX socket support
+should be removed, while the admin UNIX socket remains in clear-text
```
fmoor, line 298: Does this mean that we would undeprecate the currently deprecated admin unix socket features?

tailhook: I don't think so. It's still an appropriate way to use for auxiliary needs like resetting password, but I don't think we need it for anything else.

E.g. I don't think this is a good way for batch processing of data as disk usage time will probably dominate on the encryption time anyway.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+
+First of all, we'd want to add a new Auth method ``Certificate`` beyond
+the other two methods ``Trust`` and ``SCRAM``. The ``Certificate``
+``Auth`` entry tells the EdgeDB server which users are allowed to
+authenticate themselves using a client certificate.
+
+Then we would need a new config type to store the trusted client
+certificate fingerprints. The idea is to let the CLI generate the client
+certificates using a local CA, and the CLI tells the server to trust the
+generated certificates.
```
tailhook, line 319: We don't need storing client certificate fingerprints.

The idea of client certificates is:
1. We have a CA, certificate of which is know to the EdgeDB server
2. CA emits a client certificate, having some identifier of the user name, say common-name of `username.yourdb.example.org`
3. When client authenticates it sends the certificate from (2) to the EdgeDB server, server verifies it via certificate described in (1) [this involves client-certificate private key to encode some challenge-responses but I'm omitting the details now].

So on each authorization, client certificate is checked against certificate chain and no individual user info is stored between sesesions. (Well, there is TSL session cache, but that's a different layer of abstraction).

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
++------------+----------------+----------------+---------------------------+
+|            | Old Server     | New Server     | New Server in Compat Mode |
++============+================+================+===========================+
+| Old Client | Accessible     | Friendly Error | Accessible                |
++------------+----------------+----------------+---------------------------+
+| New Client | Friendly Error | Accessible     | Friendly Error            |
++------------+----------------+----------------+---------------------------+
+
+The EdgeDB development server (``edb server``) will provide a hidden
+option ``--no-tls`` to run the server in non-TLS compatible mode (just
```
tailhook, line 357: Can we have `--allow-non-tls` rather than `--no-tls` ? I think we should be able to detect whether the handshake was TLS one or binary protocol.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+For both the server-side and client-side (if implemented) certificate
+verification, the corresponding private keys and their passphrases are
+critical for system security. Malicious parties could use the server
+credential to start a fake but valid server, potentially being able to
+collect sensitive queries without the user knowing. And a cracker could
+use the users' credentials to access their data in the database.
+
+As the server private key passphrase is stored in the
+``credentials.json`` file in clear text, this directory needs extra
+attention for security purposes in production environments.
```
tailhook, line 386: This will not be the case if my proposal above is accepted.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+
+
+Security Implications
+=====================
+
+Enforcing TLS is supposed to be a full level-up in terms of security. It
+provides basic eavesdropping protection, and if configured properly the
+MITM protection too. This needs to be carefully documented as this RFC
+does not involve setting up a proper CA-based trust chain for MITM
+attacks.
```
tailhook, line 375: You don't need CA-based trust to prevent MITM. Veryfying server certificate is enough. So generating `credentials.json` by web UI of cloud (Aether) or providing certificate fingerprint alongside of user and password from DB Ops to Users within the company is enough for that. Client certificates are even better.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,551 @@
...
+    {
+        "port": 10732,
+        "user": "edgedb",
+        "password": "login-password-in-clear-text",
+        "database": "edgedb",
+        "tls_certdata": "-----BEGIN CERTIFICATE-----\nMIICvjCCAaagAw..."
+    }
+
+The language bindings and the REPL should load the certificate from the
+value of ``tls_certdata`` and trust only that certificate for connecting
```
elprans, line 163: I'd clarify that `tls_certdata` can specify an issuing certificate as well up to some long-lived root cert.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,551 @@
...
+
+* Python [4]_: ``set_alpn_protocols()`` and ``selected_alpn_protocol()``
+* Go [5]_: ``SupportedProtos`` and ``NegotiatedProtocol``
+* Node.js [6]_: ``ALPNProtocols`` and ``alpnProtocol``
+
+For now, the EdgeDB server will advertise two protocols in ALPN (however
+EdgeDB is not limited to only these two for future possibilities):
+
+* ``edgedb-binary``: The EdgeDB binary protocol
+* ``http``: HTTP-based protocol, including the server system API, and
```
elprans, line 254: Specifically, this would be `http/1.1`, as we might also add support for `h2` in the future too.

text/1008-tls-and-alpn.rst
fantix: Yeah sure thing - a real compatible dev server could be more handy.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+On the client side (both the language bindings and the REPL), TLS server
+certificate verification should always be enabled. In order to accept
+the self-signed certificate, the client must automatically trust the
+certificate found in ``tls_certdata`` if this key is present in the
+``credentials.json``. The generated certificate must contain a valid
+host (usually ``127.0.0.1``) in the ``subjectAltName`` extension of type
+``dNSName`` [7]_, so that the client could always keep the hostname
+check on.
```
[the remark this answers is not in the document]
fantix: Let's skip hostname check for now, and the client could have a `TLS_STRICT` mode in the future if we need the check.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+                    --tls-certfile as well.
+
+(This is exactly `ssl.SSLContext.load_cert_chain()` from Python [4]_.)
+
+The specified path will be stored in the so-called ``credentials.json``
+- a group of JSON files named after the EdgeDB instance in a well-known
+place (e.g. `~/.edgedb/credentials/` depending on the OS) that are meant
+to store credentials. The EdgeDB server instance will then load the
+certificate and private key from the designated paths on (re)start and
+reload.
```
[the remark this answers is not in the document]
fantix: Thanks for the comments and the offline discussion! To summarize, we'll move the server parameters to `metadata.json`, and leave the `credentials.json` for only client connection parameters (public key/cert only). We're not introducing CA just yet - this could come together with the future client cert verification.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,458 @@
...
+set, simply return ``b""`` in the function - it will not be invoked if
+the private key is not protected by a password.
+
+When ``--tls-certfile`` is not present and the server is in ``devmode``
+or ``testmode``, the server will use the CLI to generate a self-signed
+certificate and use it to run the TLS server for development and
+testing. The generated certificate (with its private key) should be
+written to a ``credentials.json`` file just like the regular CLI
+behavior, together with the connection parameters. For example, the
+server would run this "hidden" CLI command on start::
```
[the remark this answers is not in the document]
fantix: I've updated the RFC not to use ``credentials.json`` for private keys. But I don't think we can use the CLI-specific data dir for core EdgeDB development, and the dev server doesn't always have a stable data dir (e.g. with `--temp-dir` or `--postgres-dsn`). So I kept some of the previous design (CLI generates a new self-signed cert and send to server through stdout on startup), and updated the pub cert into ``credentials.json`` together with the new ``edgedb authenticate`` command for easier core dev.

text/1008-tls-and-alpn.rst
fantix: 👍 Thanks!  Added an open question - looking forward to more comments on this topic.

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,580 @@
...
+    --tls-certfile  Path to a single file in PEM format containing the
+                    certificate as well as any number of CA certificates
+                    needed to establish the certificate’s authenticity.
+                    If not present, a self-signed certificate will be
+                    generated and used instead (for development only!).
+    --tls-keyfile   Path to a file containing the private key. If not
+                    present, the private key will be taken from
+                    --tls-certfile as well.
+
+(This is exactly `ssl.SSLContext.load_cert_chain()` from Python [4]_.)
```
1st1, line 80:
```suggestion
(This is exactly ``ssl.SSLContext.load_cert_chain()`` from Python [4]_.)
```

Single backticks mean something else in ReST, please check if you use them elsewhere

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,580 @@
...
+
+The EdgeDB CLI will automatically generate a self-signed certificate if
+``--tls-certfile`` is not present in command ``edgedb project init`` or
+``edgedb server init`` under the data directory of that EdgeDB instance
+as described in the previous section, in the names of
+``edgedb-cert.pem`` for the certificate in PEM format, and
+``edgedb-key.pem`` for the private key (no passphrase for simplicity).
+This certificate is supposed to be used for development purposes only.
+
+Likewise, the ``metadata.json`` file will be update with the full path
```
1st1, line 138:
```suggestion
Likewise, the ``metadata.json`` file will be updated with the full path
```

text/1008-tls-and-alpn.rst
1st1: Yes, we should clarify that the support of admin UNIX socket will be left only in the CLI. All bindings will not be able to connect to UNIX sockets at all.

1st1 (approved):
LGTM. I think we should merge this and iterate in follow up PRs if necessary (I think there's only one or two yet unaddressed comments)

text/1008-tls-and-alpn.rst
```diff
@@ -0,0 +1,580 @@
...
+=====================
+
+Usually TLS just work out of the box with the default settings. But for
+special security reasons, optionally the advanced TLS settings can be
+modified in the EdgeDB config system per instance. Specifically:
+
++-------------------------+--------------------------+--------------------------------------------------------------------------+
+| EdgeDB Config           | Python SSLContext member | Possible Values                                                          |
++=========================+==========================+==========================================================================+
+| ``tls_minimum_version`` | ``minimum_version``      | ``1.0``, ``1.1``, ``1.2``, ``1.3``, ``MIN_SUPPORTED``, ``MAX_SUPPORTED`` |
```
elprans, line 294: TLS <1.2 isn't considered to be secure anymore, so I think we should have 1.2 as the minimum.

PR #42, opened 2021-07-28
rfc-1006: update instance subcommands: upgrade and reset-auth

1st1:


text/1006-simplified-cli.rst
```diff
@@ -120,11 +118,12 @@ the EdgeDB RC1 release::
...
-    self uninstall             Uninstall this tool
+    server                     Manage local installations of EdgeDB server
 
+    cli upgrade                Upgrade the 'edgedb' command to the latest version
+    cli uninstall              Uninstall the 'edgedb' command
 
-The output of ``edgedb inspect --help``::
+The output of ``edgedb list --help``::
 
   edgedb inspect [SUBCOMMAND]
```
fantix, line 128: `edgedb list` too

text/1006-simplified-cli.rst
```diff
@@ -97,19 +97,17 @@ the EdgeDB RC1 release::
...
-    server upgrade             Upgrade installations and instances
-    server list-versions       List available and installed versions of the server
-
     instance create            Initialize a new server instance
     instance status            Show statuses of all or of a matching instance
     instance start             Start an instance
     instance stop              Stop an instance
     instance restart           Restart an instance
-    instance destroy           Destroy an instance and remove the data
+    instance destroy           Destroy a local instance and remove the data
```
elprans, line 105: Why emphasis on "local" here?

text/1006-simplified-cli.rst
1st1: good catch, fixed

text/1006-simplified-cli.rst
1st1: Hm, now I'm not sure what concern I had in that editing instant. I'll drop "local".

elprans (approved):

PR #54, opened 2022-04-18
RFC 1011: Object-level security

elprans:


text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+
+
+Introspection
+=============
+
+Policies can be introspected via a new ``schema::AccessPolicy`` in the
+introspection schema that is linked from ``schema::ObjectType`` via the new
+``access_policies`` link.  The ``schema::AccessPolicy`` is exposed as follows::
+
+    type schema::AccessPolicy extending schema::AnnotationSubject {
```
msullivan, line 236: Probably an `schema::InheritingObject`, also?

text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+*how* a policy is applied.  Also, other access policies do not apply to
+``when`` expressions, but they *do* apply to ``using`` expressions.
+
+If at least one access rule is defined for a type, then all elements become
+invisible/immutable by default and must be made visible/mutable by at least
+one ``PERMIT`` rule.
+
+A ``PERMIT`` rule *adds* all objects, for which the check expression
+evaluates to true, to the set of visible objects.  ``PERMIT`` rules are
+combined using the ``OR`` operator.
```
msullivan, line 85: What's the intended behavior for `{}`? My inclination would be to treat it like `false`. Equivalently, that the rules are essentially combined with `any`

text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+    type schema::AccessPolicy extending schema::AnnotationSubject {
+      multi property access_kind -> schema::AccessKind;
+      property condition -> std::str;
+      required property action -> schema::AccessPolicyAction;
+      required property expr -> std::str;
+    };
+
+
+Implementation considerations
+=============================
```
msullivan, line 245: One implementation consideration is how to apply access policies in subtypes. I can think of two approaches:
 1. Instead of using views for types with access rules, fully expand it out into a union of all the subtypes, each with appropriate filtering rules
 2. Keep using views, and use type filtering to do the appropriate filters on subtypes

text/1011-object-level-security.rst
elprans: yes.

text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+invisible/immutable by default and must be made visible/mutable by at least
+one ``PERMIT`` rule.
+
+A ``PERMIT`` rule *adds* all objects, for which the check expression
+evaluates to true, to the set of visible objects.  ``PERMIT`` rules are
+combined using the ``OR`` operator.
+
+A ``RESTRICT`` rule restricts the set of visible objects to only objects for
+which the check expression evaluates to true.  ``RESTRICT`` rules are combined
+using the ``AND`` operator and are applied *after* all ``PERMIT`` policies.
```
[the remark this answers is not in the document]
elprans: There is a sentence above that clarifies this:

> If at least one access rule is defined for a type, then all elements become invisible/immutable by default and must be made visible/mutable by at least one ``PERMIT`` rule.

`RESTRICT` rules apply _after_ `PERMIT` and enforce exceptions from what is permitted.  This is the same logic that exists in [Postgres RLS](https://www.postgresql.org/docs/current/sql-createpolicy.html#id-1.9.3.75.6.4).

text/1011-object-level-security.rst
elprans: `{}` evalueates to `false` like it does in `filter` clauses.

text/1011-object-level-security.rst
elprans: I was thinking that the most reliable solution would be to actually _always_ bake all access policies into views themselves.

msullivan: How would they access global variables, then?

text/1011-object-level-security.rst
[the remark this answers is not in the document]
msullivan: Oh, I had misinterpreted `RESTRICT` as doing the *restriction* when the value is true---that is, inverting the meaning of the boolean.

To do some syntax quibbling, then, I think that my incorrect reading here is more natural: `restrict all using (foo)` seems to imply to me that it ought to restrict *when* `foo` holds. I understand that the terminology is made to mirror postgres, but I think using the verb `RESTRICT` instead of the adjective `RESTRICTIVE` makes this misunderstanding easier to have in our setting. I'm not sure if I have a counterproposal but I think it's worth thinking about.

text/1011-object-level-security.rst
elprans: Hm. Good point, it slipped from my mind that we're moving away from the temp table approach altogether.  Full expansion would be the best option from the standpoint of query quality, though at the cost of query text size.  A set-returning function instead of a view would be another option, but it's probably too slow and might be too opaque for the optimizer.

text/1011-object-level-security.rst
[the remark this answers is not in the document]
elprans: The previous RFC iteration had the `RESTRICT TO` spelling, would that be better?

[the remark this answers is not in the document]
msullivan: > `RESTRICT` rules apply *after* `PERMIT`

I don't really understand this claim, from a semantic point of view. `AND` is commutative.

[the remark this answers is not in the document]
msullivan: Yeah, I like `RESTRICT TO`

text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+returns a false result for an object, such object will be silently skipped
+in any ``SELECT``, ``UPDATE`` or ``DELETE`` expression that scans it.
+A ``WRITE`` policy specifies a *validity* check for new or updated objects and
+affects ``INSERT`` and ``UPDATE`` expressions: if the proposed object violates
+``<expr>``, an error is raised immediately and the query is aborted.
+
+The optional ``<condition>`` expression is evaluated for every object
+affected by the statement and the policy is applied only if the expression
+evaluates to *true*.  It is essentially equivalent to joining ``<condition>``
+with ``<expr>`` with an ``AND`` operator.  The reason for a standalone clause
```
msullivan, line 74: Is `<condition>` negated before `AND`ing with `<expr>` for `RESTRICT` queries, then? To match the intuition that the policy is only applied then?

text/1011-object-level-security.rst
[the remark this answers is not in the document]
elprans: > > `RESTRICT` rules apply _after_ `PERMIT`
> I don't really understand this claim, from a semantic point of view. `AND` is commutative.

OK, so here's how OLS works *semantically* (as in mental model):

- you start with an empty set of objects;
- you *add* to the set with `permit` rules;
- you *subtract* from the set with `restrict to` rules

Naturally, implementation-wise `and` is commutatuve in a declarative `filter` statement, but logically subtraction isn't.

text/1011-object-level-security.rst
elprans: Yeah, the wording in this paragraph is wrong, the equivalent of `when <condition>` is `not <condition> or <policy-expr>`

msullivan: It's still `AND` for `PERMIT` though, right?

elprans: No.  `PERMIT` is combined with `OR`

elprans: Oh, I misunderstood your question.  Yes, your understanding is correct.  `permit` does *not* "fail open".  In other words, if `<condition>` is false, no objects will be added to the "permitted" set.

msullivan: Looking at the examples again, the use of `when` for the `age_appropriate` access policy seems off. That rule probably ought to be `restrict to`? (Which then would require a catch all `permit` rule?)

text/1011-object-level-security.rst
```diff
@@ -0,0 +1,320 @@
...
+-------------------
+
+Alter the definition of an access control policy.
+
+Required capabilities: DDL.
+
+Synopsis::
+
+    ALTER OBJECT TYPE <type-name> "{"
+        ALTER ACCESS POLICY <name>
```
msullivan, line 153: I'm not sure I understand what is intended here, with all of the arguments to `CREATE` being recapitulated in `ALTER`? Typically those are omitted and we modify the fields with subcommands, right?

Is the intent that since access policies don't have data and can't be referenced in any way, that they don't even really need `ALTER`, and we can just generate `drop`/`create` for actual changes? (Though I'd still expect ALTER to only have the name, not all the extra stuff?)

elprans: This is meant to indicate that any part of the access policy can be altered by re-stating it in `alter`.

msullivan: Aha. Is the idea that the restatement is mandatory? And so, for example, omitting `WHEN` would drop it?

elprans: I think it would be more consistent if we didn't actually *require* restatement.  Perhaps `drop when`?

msullivan: Yeah that could work

Issue #55, opened 2022-04-26
Pre-RFC: Range types


vpetrovykh:
## Range types

Range types can exist as a sort of a scalar value, akin to arrays. Although they represent a set of values, but the way we manipulate these values is different from a pure EdgeQL set. It's a special case that is a little more structured.

Other programming languages seem to mostly use range types to loop over indexes. This biases the way ranges are defined. In EdgeDB range types are expected to be used in constraints or even as values unto themselves.

### Base types

It seems like we need the following abstract types:
* `anypoint` - any scalar that has "points" which can be used to construct a range (like `int64` and `float64`, but not, say, `json` and `str`)
* `anydiscrete extending anypoint` - any scalar that conceptually represents discrete values (like `int64` or `date`)
* `anycontiguous extending anypoint` - any scalar that conceptually represents contiguous values (like `float64` or `datetime`)
* range itself would be typed similarly to arrays: `range<sometype>` where `sometype` is a concrete type extending `anypoint`.

### Boundaries

A range value should have an upper and lower bound and an indicator for each of them whether they are included in the values the range represents.

It should be possible to omit one or both of the boundaries to represent an open-ended interval. The `{}` seems like a reasonable representation of such an omitted boundary. It should be an error to omit the boundary while indicating that it is inclusive.

### Constructors

One option is to have a special literal to construct a range. A particularly neat format can draw inspiration from the way ranges are indicated in math by using round and square brackets to indicate boundary inclusivity. E.g. `[5..10)` or `(..-10]`. The upside is that this format is familiar from math. The downside is that this messes with any parentheses matching processing (auto-complete, highlighting, etc.)

Another option is to have matching parentheses and use an extra symbol to indicate inclusivity. E.g. `(=5..10)` or `(..=-10)`. It seems that exclusive boundary is the more natural default for this approach, largely because it works naturally with omitted boundaries. The upside of this format is that it plays nice with parentheses matching. The downside is that it's hard to search for and it's not necessarily intuitive when seeing it.

Finally, we can have a function constructor. To make things succinct the signature should be:
```
function range(
    lower: optional anypoint = {}, upper: optional anypoint = {},
    inclow: bool = true, inchigh: bool = false
) -> range<anypoint>
```
Basically, we want to be able to use positional arguments for brevity. We also want the common case of using ranges for some kind of iteration to only need the boundaries and have the inclusivity match how we use indexes in slicing.

The constructor function is compatible with also having a dedicated literal, so it can be implemented first.

[...]

PR #57, opened 2022-05-04
RFC 1012: Range types

vpetrovykh:
Initial proposal to add range types to EdgeQL.

### DIFF
--- a/text/1012-range-types.rst
+++ b/text/1012-range-types.rst
@@ -1,0 +1,310 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2022-05-02
+    Authors: Victor Petrovykh <victor@edgedb.com>
+    RFC PR: `edgedb/rfcs#0054 <https://github.com/edgedb/rfcs/pull/57>`_
+
+=====================
+RFC 1012: Range types
+=====================
+
+This RFC proposes adding ``range`` and ``multirange`` scalar types to EdgeQL.
+The ``range`` type represents an *interval* of values. While ``multirange``
+type represents a set of non-overlapping ranges.
+
+
+Motivation
+==========
+
+Having a scalar type representing intervals would allow defining a consistent
+way of working with them, such as determining whether two intervals overlap,
+whether one interval is contained in the other, etc. The intervals could also
+be used in constraints. A simple usage would be to define a single constraint
+using an interval instead of using ``min_value`` and ``max_value``
+constraints. More advanced usage could be defining a variant of the
+``exclusive`` constraint which ensures non-overlapping interval property for
+some objects.
+
+Another usage of intervals is in generating sequences of values to be used
+directly or to serve as an iterator set in a ``for`` loop.
+
+
+Specification
+=============
+
+Range types would be similar to ``array`` in that they represent a collection
+of values of a particular type. Since intervals in Postgres are not completely
+arbitrary, we will only be able to define ``range`` and ``multirange`` over
+specific scalar types.
+
+The base scalar types that ``range`` and ``multirange`` can work with:
+* ``anypoint`` - any scalar that has "points" which can be used to construct
+  a range (like ``int64`` and ``float64``, but not, say, ``json`` and ``str``)
+* a helper type expression ``increment<type>``, which is syntax sugar for
+  looking at the return type of the operator ``-(type, type)``. It serves to
+  identify the appropriate type for "stepping" through values inside an
+  interval.
+* ``anydiscrete extending anypoint`` - any scalar that conceptually
+  represents discrete values (like ``int64`` or ``date``)
+* ``anycontiguous extending anypoint`` - any scalar that conceptually
+  represents contiguous values (like ``float64`` or ``datetime``)
+
+The interval types themselves would be defined similar to arrays:
+* ``range<anypoint>`` represents intervals over some concrete ``anypoint``
+  sub-type
+* ``multirange<anypoint>`` represents a set of non-overlapping ``range``
+  values over the corresponding concrete ``anypoint`` sub-type
+
+
+Boundaries
+----------
+
+A ``range`` value should have an upper and lower bound and an indicator for
+each of them whether they are included in the values the ``range`` represents.
+
+It should be possible to omit one or both of the boundaries to represent an
+open-ended interval. The ``{}`` will be used to represent the omitted
+boundary. It should be an error to omit the boundary, but indicate that it is
+inclusive.
+
+
+Constructors
+------------
+
+We will use a function constructor for creating new ``range`` values::
+
+  function range(
+      lower: optional anypoint = {},
+      upper: optional anypoint = {},
+      named only inc_lower: bool = true,
+      named only inc_upper: bool = false
+  ) -> range<anypoint>
+
+The default values are chosen so as to minimize the amount of keystrokes
+required to define ranges for common use-cases such as for iterators.
+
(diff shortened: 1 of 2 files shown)


vpetrovykh:
This is an RFC draft based on #55.

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Accessing range components
+--------------------------
+
+Several functions will be used to access ``range`` components::
+
+  function get_range_upper(range<anypoint>) -> optional anypoint
+  function get_range_lower(range<anypoint>) -> optional anypoint
+  function is_inclusive_lower(range<anypoint>) -> bool
+  function is_inclusive_upper(range<anypoint>) -> bool
```
elprans, line 110:
```suggestion
  function range_get_upper(range<anypoint>) -> optional anypoint
  function range_get_lower(range<anypoint>) -> optional anypoint
  function range_is_inclusive_lower(range<anypoint>) -> bool
  function range_is_inclusive_upper(range<anypoint>) -> bool
```

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+
+
+Accessing multirange components
+-------------------------------
+
+There will be a function for extracting sub-ranges from a multirange::
+
+  function get_subranges(multirange<anypoint>) -> set of range<anypoint>
```
elprans, line 118: Maybe

```suggestion
  function multirange_unpack(multirange<anypoint>) -> set of range<anypoint>
```

?

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+
+By default this function will return the ranges ordered by boundaries
+fromlowest to highest. This is the order of definition and Postgres by default
```
elprans, line 121:
```suggestion
from lowest to highest. This is the order of definition and Postgres by default
```

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Helper functions
+----------------
+
+In addition to the functions mentioned above there must be a way to generate
+an ordered set of values from a ``range`` or ``multirange``::
+
+  function range_unpack(
+      val: range<anydiscrete>,
+      step: anydiscrete = DEFAULT_STEP
```
elprans, line 133: Would be easier to add overloads instead of sentinel `DEFAULT_STEP`.

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+
+  function range_unpack(
+      val: multirange<anydiscrete>,
+      step: anydiscrete = DEFAULT_STEP
+  ) -> set of anydiscrete
+
+
+  function range_unpack(
+      val: range<anycontiguous>,
+      step: anycontiguous
```
elprans, line 144: We would have to invent another type: `increment<anycontiguous>` (and `increment<anydiscrete>`), which would internally resolve into the appropriate increment type of a given point type, i.e. `duration` for `datetime`.

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+      haystack: range<anypoint>,
+      needle: range<anypoint>
+  ) -> bool
+
+  function overlaps(
+      haystack: multirange<anypoint>,
+      needle: multirange<anypoint>
+  ) -> bool
+
+It should be noted that if ``contains(A, B) = true`` then ``overlaps(A, B) =
```
elprans, line 191: Postgres defines a bunch more functions for ranges, such as relative disposition (e.g. "strictly to the left of", "strictly to the right of", etc.).  I think we should reflect those as well.

text/1012-range-types.rst
```diff
@@ -0,0 +1,277 @@
...
+true``.
+
+
+Operators
+---------
+
+The operators ``+`` and ``-`` should be defined for various combinations of
+``range`` and ``multirange``. The semantics of ``+`` is a "union", whereas the
+``-`` allows excluding some values from an interval. Both of these operators
```
elprans, line 200: What about the intersection of ranges?

text/1012-range-types.rst
vpetrovykh: Yeah, good point about our naming scheme.

text/1012-range-types.rst
vpetrovykh: This is very long name... maybe `to_range(multirange<anypoint>)` instead? It's a conversion function of sorts.

text/1012-range-types.rst
vpetrovykh: Fixed typo.

text/1012-range-types.rst
vpetrovykh: Yes, you're right, I was sloppy here and I should explain that we want overloads.

text/1012-range-types.rst
vpetrovykh: I don't think that inventing another type makes any sense here because for `local_date`, `local_time`, and `datetime` the step is `duration`, but it's also a valid step for `duration` itself. Trying to link them by marking them with abstract types doesn't seem to work because abstract types don't let us see the valid pairs.

Date and time scalars seem to be the only current exception, but if we ever have another one, generalizing this mechanism would be near impossible. On the other hand, encoding this feature as a particular overload for `range_unpack` is clear and future-proof.

text/1012-range-types.rst
vpetrovykh: Yeah, you're right.

text/1012-range-types.rst
vpetrovykh: Hmm, yeah I missed an important bit here. We need to define `>`, `>=`, `<`, and `<=` for everything in EdgeQL (because everything must be orderable). So I'll add that to "operators". Meanwhile we'll have functions for "strictly to the left/right" and "does not extend to the left/right of".

The question is what to call these? They are wordy and I'd like to avoid full sentences as function names.

text/1012-range-types.rst
elprans: `increment<type>` is not "another type", it's an alias for the return type of operator `-(type, type)`, i.e. it needs no actual formal implementation aside from a bit of schema magic.

text/1012-range-types.rst
elprans: `to_range` implies that a single range would be returned, `to_ranges` would be better.

vpetrovykh: Fair enough.

text/1012-range-types.rst
vpetrovykh: Wait, are you suggesting that `increment<type>` is a type expression that isn't independently defined, but instead literally looks up the operator `-(type, type)` and returns the declared return type of that operator?

And the main reason for this thing even existing is to be used in formal generic function definitions like `range_unpack(val: range<anypoint>, step: increment<anypoint>)`? What's the benefit? We would still implement the specific overloaded versions for each specific range type, I think.

If it is independently defined, then it would place the burden of reconciling it with the operator, so maybe not a great option.

elprans: > but instead literally looks up the operator `-(type, type)` and returns the declared return type of that operator?

This.

> We would still implement the specific overloaded versions for each specific range type, I think.

If we implement the "magic" type then we wouldn't necessarily need to implement the overloads, because the compiler can just assume that all descendants of `anypoint` implement the necessary generator.

vpetrovykh: Added `increment<type>` to the RFC.

PR #59, opened 2022-06-07
RFC 1013: Date/time arithmetic.

vpetrovykh:
Add subtraction operator `-` for all "local" date/time scalars.

Add helper funcitons for normalizing `relative_duration` and
for extracting its various components.


text/1013-datetime-arithmetic.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Specification
+=============
+
+The ``local_datetime`` will introduce the following subtraction operator::
+
+  CREATE INFIX OPERATOR
+  std::`-` (
+    l: cal::local_datetime,
+    r: std::local_datetime
```
elprans, line 94:
```suggestion
    r: cal::local_datetime
```

text/1013-datetime-arithmetic.rst
```diff
@@ -0,0 +1,216 @@
...
+          SELECT ("l" - "r")::edgedb.relative_duration_t
+      $$;
+  };
+
+The result is expected to be a ``relative_duration`` in days and
+hours/minutes/seconds. This is the most precise format as ``local_datetime``
+cannot account for DST and so 1 day and 24 hours are always equivalent
+durations in this context.
+
+The ``local_datetime`` will introduce the following subtraction operator::
```
elprans, line 106:
```suggestion
The ``local_time`` will introduce the following subtraction operator::
```

elprans (approved):

text/1013-datetime-arithmetic.rst
```diff
@@ -0,0 +1,226 @@
...
+std::duration_get(dur: std::duration, el: std::str) -> std::float64
+{
+  ...
+};
+
+The components avaialable for extraction from ``relative_duration`` are:
+*millenium*, *century*, *decade*, *year*, *quarter*, *month*, *day*, *hour*,
+*minutes*, *seconds*, *milliseconds*, *microseconds*, and *totalseconds*. The
+*totalseconds* converts a ``relative_duration`` to seconds represented as a
+``float64`` value.
```
tailhook, line 196: We need to document what does `duration_get` for `std::duration`. It never returns anything bigger than `hours`, right? Is asking for that is an error or zero result?

Also why we don't use enum here, is it because of performance issue?

text/1013-datetime-arithmetic.rst
```diff
@@ -0,0 +1,226 @@
...
+There are no security implications.
+
+
+Rejected Alternative Ideas
+==========================
+
+We specifically rejected the option of ``-`` operating on two ``local_date``
+values producing a ``relative_duration`` by default. In practice, this
+operation would produce ``relative_duration`` expressed using exclusively days
+anyways. The main difference then would be that in order to use the result of
```
tailhook, line 219: Why does it produce days only and not months or years? Is it artifact of the postgres implementation?

tailhook: Also even if it does like that, I'm still thinking that  `local_date - local_date = days` isn't obvious. One could assume it produces seconds or duration type as well depending on their background. So returning `relative_duration` with days only makes a lot of sense making code like `duration_get(date1 - date2,
'days')` self-documenting.

vpetrovykh: > Why does it produce days only and not months or years? Is it artifact of the postgres implementation?

Because for date arithmetic `days` are exact, whereas months/years/etc. are not. It is also the case that Postgres implements it this way.

Perhaps we want to expose `age` function from Postgres that performs the subtraction exactly and returns the result using days, months, years, etc.

> So returning `relative_duration` with days only makes a lot of sense making code like `duration_get(date1 - date2, 'days')` self-documenting.

There's a performance penalty if you actually want days as an integer: `<int64>duration_get(date1 - date2, 'days')`. It's also pretty bulky for a common use-case.

Another reason relates to future plans for `range_unpack`. We want to be able to generate a bunch of dates from a range. By default we want to step by 1 day, but in principle we also want to allow other steps, like 2 days, 7 days, etc. The key here is that it's always a whole number of days, not fractional. So we've had this idea that it's really the `-` operator that defines the type of the step (as it does for all other types - because the step is in whatever units the "difference" is measured). If we use `relative_duration` for step instead we need to either discard the fractional part (like it happens for performing `+`) or forbid it outright making it an error. Both seem rather awkward. So these thoughts led us to the idea that actually `local_date` is much more like a special kind of integer than it is like `datetime`.

text/1013-datetime-arithmetic.rst
```diff
@@ -0,0 +1,226 @@
...
+cal::normalize_relative_duration(dur: cal::relative_duration)
+  -> cal::relative_duration
+{
+    USING SQL FUNCTION 'justify_interval';
+};
+
+The ``normalize_relative_hours`` converts 24 hrs chunks into days.
+The ``normalize_relative_days`` converts days into months assuming that 1 month = 30 days.
+The ``normalize_relative_duration`` is the equivalent of calling
+``normalize_relative_days(normalize_relative_hours(x))``.
```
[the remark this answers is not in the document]
elprans: I actually think that the original "justify" name is probably better here, though whatever we choose it's not going to be 100% descriptive unless we go the super-verbose route.

text/1013-datetime-arithmetic.rst
elprans: Yeah, the analogy here is integers vs floating point types.  `relative_duration` is fractional, so it'd be like returning a `float64` from `int64 - int64`, which is loss of type fidelity.  As a compromise we can consider adding a `relative_duration` subdomain that allows only `days` and is implicitly castable to `relative_duration` and can also be cast from/to ints.  This way you can still get nice typing and output but without the ambiguity of `relative_duration` or weirdness of pure `int64`.

text/1013-datetime-arithmetic.rst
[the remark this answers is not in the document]
tailhook: > though whatever we choose it's not going to be 100% descriptive unless we go the super-verbose route.

My problem is mostly that `normalize_relative_days` is not clear whether it normalizes up (i.e. days to months and years) or down (i.e. hours to days). Also technically `normalize_relative_hours` normalizes "microseconds" instead (i.e. there is no `hours` counter in the internal structure).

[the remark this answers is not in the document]
tailhook: Also, I like the precedent that [systemd set](https://www.freedesktop.org/software/systemd/man/systemd.time.html) is that months are counted as `30.44` days and years are set to `365.25` days. For big durations, it's much more accurate for human presentation.

On the other hand, "normalizing" relative datetime that had just days and getting non-zero hours in the "normalized" variant is also unexpected for most users. So it's more useful for creating rather than formatting and for durations rather than relative durations.

And then, as you have said, years are also useful and they will be calculated from months, but their duration will be `360` days with the semantics defined in the current spec.

What I'm trying to say is that there is no generally agreed "normalization" operation. 

So we might want few formatting and/or truncation operations instead, with customizable ways to convert units.

text/1013-datetime-arithmetic.rst
tailhook: > There's a performance penalty if you actually want days as an integer: `<int64>duration_get(date1 - date2, 'days')`. It's also pretty bulky for a common use-case.

I think we can instruct compiler to optimize this in most cases. While it's bulky, I'm pretty sure it's not so common to optimize brevity over clarity.

-----

On the other hand, I somewhat agree with the "minus defines type of the step" argument.

Will I still be able to step by month or year with this definition? Is it applied only for default step, or does this mean that integer days is the only step type I can use?

text/1013-datetime-arithmetic.rst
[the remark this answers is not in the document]
vpetrovykh: I've added an overload for `duration_truncate` for `relative_duration` to this PR. So with some combination of `duration_get`, `duration_truncate` and possibly some normalizations you can try to massage the `relative_duration` into a reasonable presentation.

At the end of the day, if you're trying to normalize `relative_duration` you've already lost your reference point and will inevitably make some approximations, so that process is like calling `round`. For more accurate `relative_duration` values we need to expose `age` (possibly under some other name), which performs the arithmetic as accurately as possible for the given 2 datetimes.

text/1013-datetime-arithmetic.rst
vpetrovykh: Updated the RFC

text/1013-datetime-arithmetic.rst
vpetrovykh: > Will I still be able to step by month or year with this definition? Is it applied only for default step, or does this mean that integer days is the only step type I can use?

Good point.

I think we should go with the suggestion by @elprans: introduce a new subtype of `relative_duration` that does not have the "fractional day" components. We can call it `date_duration`.

text/1013-datetime-arithmetic.rst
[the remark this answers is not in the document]
tailhook: > if you're trying to normalize `relative_duration` you've already lost your reference point and will inevitably make some approximations, so that process is like calling `round`.

Yes. But there is a difference. It's fine if `date + normalize_relative_duration(x)`
comparing to `date + x` is off by hours (or maybe days on large numbers). But with the semantics above, it will be off by two years if `x > "55000 days"`.

I you really want something like rounding behavior then month length should be about `30.44` days and rounding is used to throw away fractional days. But I'm not sure if this makes sense if fractional days are already present in the input.

vpetrovykh:
I've added `relative_delta` as a way to expose `age` from Postgres as it represents a very useful wayof performing subtraction of `local_date`/`local_datetime` values.

I've also updated the int/float analogy for the new arithmetic. This potentially adds a backwards-incompatible change for `+`.

PR #61, opened 2022-07-21
Update RFC 1011.

vpetrovykh:
Add the concept of `access group` in addition to `access policy` as a
way to address code duplication and boilerplate.


vpetrovykh:
Adding `access group` attempts to provide the grouping functionality that we seem to want for access policies without affecting the existing DDL too much. Additionally this mechanism is more general than allowing anonymous sub-policies under a single `access policy`, because it also allows grouping together of policies affecting the same type of access.

1st1:
Looks good to me after a quick glance. But since this is a new capability that we would be shipping in a future release I think it makes sense to describe it in a *separate* RFC. This was we can mark this as *final* and the reader will have a clear understanding of what's shipped in 2.0 and what will be implemented later.

vpetrovykh:
OK, I'll split it into its own RFC. The general idea here is to keep our current `access policy` approach and propose a way to build on top of it so that we don't have to do last minute changes to 2.0 release.

Once I submit a new RFC I'll cancel this PR, but for now we can gather comments here if there are more.

vpetrovykh:
I've pushed #63 to address adding the grouping mechanism. However, I've updated the current PR to have one modification: remove `when` clause. If we kill it, the sooner we do that the better.

PR #63, opened 2022-07-22
RFC 1014: Add access policy groups.

vpetrovykh:
Add a mechanism for grouping access policies together to enable code
re-use and clearer semantics.


1st1 (changes requested):
Hm, the asymetry between `json_get` and `json_set` worries me: in `json_get` the path arg is variadic, but here it's an array. I get why it's designed this way, but I don't like this.

1st1 (commented):
Sorry, please disregard my last review comment.

PR #64, opened 2022-08-15
RFC 1015: Full text search

vpetrovykh:
A draft proposal of the full-text search funcitonality for EdgeDB.

### DIFF
--- a/text/1015-full-text-search.rst
+++ b/text/1015-full-text-search.rst
@@ -1,0 +1,431 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2022-08-12
+    Authors: Victor Petrovykh <victor@edgedb.com>
+
+==========================
+RFC 1015: Full text search
+==========================
+
+This RFC outlines a generic approach to full-text search (also referred to as
+FTS in this document) functionality, keeping the core features the same across
+various specific implementations.
+
+
+Motivation
+==========
+
+Advanced text search functionality is important when dealing with text fields
+containing natural human language. When filtering such fields simpler tools
+such as ``LIKE`` or regular expressions are not flexible enough. Full-text
+search tools can be built-in (based on Postgres FTS capability) or external
+(e.g. Elasticsearch). External tools would likely be implemented in the form
+of extensions. However, we want the EdgeQL queries that use FTS to be the
+same, regardless of how the search is implemented. This means that this
+functionality should use the same built-in functions and even the query
+sub-language should be the same. Additionally, we want to make it easier to
+look for entire *objects* matching a particular search query based on all/some
+of the ``str`` fields of that object type.
+
+How Postgres does it?
+---------------------
+
+It is useful to consider how Postgres handles full-text search (generally
+other FTS implementations also support similar features, so this is useful
+baseline). Postgres introduces two types for this purpose: ``tsvector`` and
+``tsquery``.
+
+The ``tsvector`` type stores a processed version of the text. It basically
+contains information about the tokens making up the text well as their
+positions. This is the version of the text that is actually targeted by the
+FTS. Postgres also provides a few different ways to convert plain text into a
+``tsvector``. There's also a way to create an index that converts a text
+column into a ``tsvector`` and enables fast searches. Functions that convert
+text into ``tsvector`` can also take an argument that determines which parser
+will be used for the conversion.
+
+The ``tsquery`` type encapsulates the search conditions, such as which lexemes
+should or must appear and their relative order. Postgres also has several ways
+of converting a string into a ``tsquery``.
+
+Postgres provides a way to assign relative weights to different lexemes in the
+``tsvector`` which then can be used to rank FTS matches.
+
+
+Specification
+=============
+
+There's a lot of functionality involved in supporting FTS, but it's a fairly
+niche thing, so it makes sense to create a ``module fts`` for all of the FTS
+tools.
+
+In EdgeDB we should avoid having an explicit "document" scalar like
+``tsvector``. Instead we can rely on various parameters, annotations and
+indexes to specify how the plain text ``str`` should be parsed into tokens.
+This will allow us to keep the queries involving FTS mostly implementation
+agnostic. Some FTS backends may have functions specific to them, but the goal
+is to provide good coverage of common FTS functionality.
+
+In order to simplify FTS query language, we will use a ``str`` to represent
+queries rather than introducing a new scalar type. This string is expected to
+contain FTS search queries in a format similar to how they might be typed into
+a search field.
+
+The searching functionality can be split along two axis:
+1) what FTS operates on (``str`` vs ``Object``)
+2) what FTS returns (``bool`` vs some ranked results)
+
+The task of testing whether a document matches an FTS query would be performed
+by the following built-in functions::
+
+  function fts::test(doc: str, query: str) -> bool
+  function fts::test(doc: Object, query: str) -> bool
+
+The input can be either a ``str`` or an entire ``Object``. The object type
+must have some ``str`` properties, however. Trying to perform FTS on a type
+without ``str`` properties is an error. This function is generally intended to
+be used in ``filter`` clauses in cases where there's no need for hinting what
+matched, but the results are simple enough to be self-explanatory.
+
+The ranked results need to be rich objects with several relevant properties,
+akin to free objects used for returning ``group`` results. The difference is
+that we cannot use a ``FreeObject`` as a return type of a function, we need a
+specialized type for that: ``FTSResult``. This type is defined as follows::
+
+  type fts::FTSResult {
+    property rank -> float64 {
+      default := 0;
+    }
+    property highlights -> array<str> {
+      default := [];
+    }
+  }
+
+If ranking was requested, then the value will be written in ``rank``.
+
+If highlighting the matched text was requested, then the appropriate value
+will be written in ``highlights`` array. The array is needed in case the
+matches are highlighted separately (e.g. only the matched words are returned,
+omitting the surrounding text).
+
+These enhanced FTS matches would be performed by the following built-in
+functions::
+
+  function fts::match(
+    doc: str,
+    query: str,
+    rank_opts: optional str = 'default',
+    highlight_opts: optional str = {}
+  ) -> FTSResult
+
+  function fts::match(
+    doc: Object,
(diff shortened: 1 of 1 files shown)


text/1015-full-text-search.rst
```diff
@@ -0,0 +1,368 @@
...
+however either one of them can take the value of ``{}`` with the meaning that
+ranking/highlighting of the results is not necessary. Both of them should also
+always accept ``default`` as a valid argument indicating that
+ranking/highlighting of the results should be done in some basic manner
+(typically using some faster method).
+
+We also want some way of telling when one ``ftsearch`` value is encountered as
+part of another::
+
+  function contains(haystack: ftsearch, needle: ftsearch) -> bool
```
elprans, line 139: Hm, can we use `contains` instead of `fts_test`?

vpetrovykh: I was thinking back to regexp test/match duality here. I'm not hating `contians` here, though. If there is a vote or two more for it, we could change that.

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,372 @@
...
+without ``str`` properties is an error. This function is generally intended to
+be used in ``filter`` clauses in cases where there's no need for hinting what
+matched, but the results are simple enough to be self-explanatory.
+
+The ranked results need to be rich objects with several relevant properties,
+akin to free objects used for returning ``group`` results. The difference is
+that we cannot use a ``FreeObject`` as a return type of a function, we need a
+specialized type for that: ``FTSResult``. This type is defined as follows::
+
+  type FTSResult {
```
elprans, line 97: Not a huge fan of this composition.  Realistically, can we avoid including the input object/string in the result?  In practice the input would be a singleton (e.g. in a computed or in `filter`), so perhaps the result can be a named tuple instead: `<rank: float64, highlights: array<str>>`.

Example:

```
select Object
order by fts_match(Object).rank
limit 10
```

elprans: Another point against a concrete type is that semantically those should be valid objects, i.e. they should exist and be queryable.

vpetrovykh: Either one of the `rank` and especially `highlighs` (because highlighting is expensive and not always needed) can be empty. We need a placeholder for those cases, so we can use `0` for rank and `''` for no highlights.

elprans: Sure, nothing wrong with having "zero" value here instead of an empty set

vpetrovykh: Removed the copies of hits from `fts_match` results. Added defaults for `rank` and `highlights`.

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+of the specif FTS backend (Postgres, Elasticsearch, Lucene, etc.).
+
+The searching functionality can be split along two axis:
+1) what FTS operates on (``str`` vs ``Object``)
+2) what FTS returns (``bool`` vs some ranked results)
+
+The task of testing whether a document matches an FTS query would be performed
+by the following built-in functions::
+
+  function fts_test(doc: str, query: ftsearch) -> bool
```
1st1, line 83: fts::test

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+
+In EdgeDB we should avoid having an explicit "document" scalar like
+``tsvector``. Instead we can rely on various parameters, annotations and
+indexes to specify how the plain text ``str`` should be parsed into tokens.
+This will allow us to keep the queries involving FTS mostly implementation
+agnostic. Some FTS backends may have functions specific to them, but the goal
+is to provide good coverage of common FTS functionality.
+
+In order to enable various way of manipulating the FTS query this proposal
+introduces ``ftsearch`` scalar type. It is meant to be analogous to
```
1st1, line 72: fts::search?

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+without ``str`` properties is an error. This function is generally intended to
+be used in ``filter`` clauses in cases where there's no need for hinting what
+matched, but the results are simple enough to be self-explanatory.
+
+The ranked results need to be rich objects with several relevant properties,
+akin to free objects used for returning ``group`` results. The difference is
+that we cannot use a ``FreeObject`` as a return type of a function, we need a
+specialized type for that: ``FTSResult``. This type is defined as follows::
+
+  type FTSResult {
```
1st1, line 97: fts::Result

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+
+If highlighting the matched text was requested, then the appropriate value
+will be written in ``highlights`` array. The array is needed in case the
+matches are highlighted separately (e.g. only the matched words are returned,
+omitting the surrounding text).
+
+These enhanced FTS matches would be performed by the following built-in
+functions::
+
+  function fts_match(
```
1st1, line 116: fts::match

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+- match a phrase exactly
+- match a prefix (prefix + wildcard).
+
+There are three broad approaches to implementing ``ftsearch`` operations:
+
+1) The operations are encoded entirely in the ``str`` that's being cast or fed
+   into ``to_ftsearch``.
+
+   By default, punctuation and other symbols are ignored when performing FTS,
+   so we can safely use some symbols as part of the query language.
```
1st1, line 172: You need to add examples to (1) (2) and (3)

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+
+   We could follow the path of trying to minimize the number of different
+   symbols/keywords EdgeQL uses and re-use some existing operators as
+   follows:
+
+   - ``and`` for indicating that both operands *must* match
+   - ``or`` for indicating that at least one operand *must* match
+   - ``not`` for indicating the the operand *must not* be in the result
+   - ``++`` for indicating that the operands must match in a specific order
+     one immediately after the other
```
1st1, line 188: Again, without an example this is hard to follow.

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+   this use-case and not visually ambiguous.
+
+3) We could instead use functions to perform operations on ``ftsearch``
+   values.
+
+   In that case we still probably want to use ``( )`` for grouping, though.
+
+   As for the rest of the operations we can use:
+
+   - ``and(a: ftsearch, b: ftsearch)``
```
1st1, line 218: `and` is a reserved keyword in EdgeQL

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+Wildcards would probably be part of the individual strings being converted to
+``ftsearch`` and can be indicated by ``*``. So ``<ftsearch>'wild*'`` would
+represent a FTS query matching any word that starts with "wild".
+
+As for other operations, we want to provide a reasonable ``str``
+representation of the ``ftsearch`` values. This means that we should support
+casting between ``str`` and ``ftsearch`` as well as *converter* functions,
+which might accommodate slightly different formats of the FTS strings::
+
+  function to_ftsearch(query: str, option: str = 'edgedb') -> ftsearch
```
1st1, line 239: fts::to_search

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+use a special annotation on the properties to change its weight. The proposal
+is to add a ``fts_weight`` annotation for this purpose.
+
+For example::
+
+  type Post {
+    required property title -> str;
+
+    required property content -> str {
+      annotation fts_weight := '0.5';
```
1st1, line 370: fts::weight

text/1015-full-text-search.rst
tailhook: Is `haystack: ftseach` a typo? Or what the idea behind checking whether a search term contains another search term?

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,479 @@
...
+There are three broad approaches to implementing ``ftsearch`` operations:
+
+1) The operations are encoded entirely in the ``str`` that's being cast or fed
+   into ``to_ftsearch``.
+
+   By default, punctuation and other symbols are ignored when performing FTS,
+   so we can safely use some symbols as part of the query language.
+
+2) Another option is to add operators to EdgeQL for the purpose of
+   manipulating the ``ftsearch`` values.
```
tailhook, line 175: I think this is very "dangerous" thing to do. If people will start using query builder to compose edgedb queries based on user queries, this can quickly become explosion of the query cache (both on EdgeDB and client side).

tailhook:
I think tsquery-, and tsvector-like types should be discussed in the spec even if rejected.

The idea is that query (and/or content) could be parsed on the client in any client-specific way is still very solid. I think the biggest issue here is that to make it efficient for postgres, the representation of that has to be postgres-specific, right?

1st1:
> I think the biggest issue here is that to make it efficient for postgres, the representation of that has to be postgres-specific, right?

We can make them using JSON in runtime and casting to postgres-specific underlying types when needed.

text/1015-full-text-search.rst
```diff
@@ -0,0 +1,404 @@
...
+  type Post {
+    annotation fts::language := 'English';
+
+    required property title -> str;
+    index on (.title) using fts {
+      annotation fts::index_prefixes := 'true';
+    }
+
+    required property content -> str;
+    index on (.content) using fts;
```
1st1, line 249: fts::index?

tailhook:
> > I think the biggest issue here is that to make it efficient for postgres, the representation of that has to be postgres-specific, right?
> 
> We can make them using JSON in runtime and casting to postgres-specific underlying types when needed.

The problem is that Elasticsearch's json is something like this:
```
{
  "query": {
    "query_string": {
      "query": "(new york city) OR (big apple)",
      "default_field": "content"
    }
  }
}
```
But in postgres it should be like this, I think:
``` 
(new:A & york:A & city:A) OR (big:A & apple:A)
```
I.e. its not just get the query from inside the JSON, but also operators in text are different and how fields are matched is different, etc. And this is just the simplest query for elastic.

elprans (approved):

PR #66, opened 2022-10-14
RFC 1017: Abstract index

vpetrovykh:
Add a way to define different indexes in extensions and built-in modules. Also provide a way for users to define custom indexes extending existing ones with some of the parameters set to specific values.

### DIFF
--- a/text/1017-abstract-index.rst
+++ b/text/1017-abstract-index.rst
@@ -1,0 +1,159 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2022-10-07
+    Authors: Victor Petrovykh <victor@edgedb.com>
+
+========================
+RFC 1017: Abstract index
+========================
+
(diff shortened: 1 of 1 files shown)


msullivan:
Could you update with an example of the parameter overriding mechanism in use? I'm having a little trouble fully visualizing it

tailhook:
Yes, this definitely needs some examples.

Also my primary concern is what will be defined here. Because there was some discussion in slack about making address of the full-text search index server a part of the (abstract) index definition. Which is not something that would work well (server address must be a part of the configuration and differs on production and staging servers).

vpetrovykh:

This portion is now trimmed and mostly concerns itself with simple abstract indexes exposing built-in Postgres index types. There's a question of how we define which types are indexable as well as where and how the SQL that is needed to define an index should be specified.

msullivan (approved):

PR #68, opened 2022-11-16
Update CLI release workflow

tailhook:
* We don't do `releases/` branch
* On the other hand we do `stable/2.2.x` branch
* Also uses more explicit tag and branch names. I think it's better in release flow, so that you don't accidentally push anything extra.
* Other minor changes

### DIFF
--- a/text/0002-edgedb-release-process.rst
+++ b/text/0002-edgedb-release-process.rst
@@ -100,31 +100,46 @@ Publishing the binaries
 CLI
 ---
 
-1. Create a release branch:
+1. Create a release branch if not exists (assuming you release 2.2.0):
 
-   - ``git switch -c releases/1.0a6``
+   - ``git switch -c stable/2.2.x``
 
-   - update version in ``Cargo.toml`` and run ``cargo check``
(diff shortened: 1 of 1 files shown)


tailhook:
Thing I'm not sure about is how do we maintain `stable/2.x` branch:
1. Should we do `git push +stable/2.2.x:stable/2.x` (i.e. force push to replace branch)?
2. Or should we do the merge?
3. Or should we do `git merge -s ours` to avoid conflicts but maintain history and pull'ability of the branch?

elprans:
> Thing I'm not sure about is how do we maintain `stable/2.x` branch:

We cherry-pick commits normally.

elprans (approved):

tailhook:
> > Thing I'm not sure about is how do we maintain `stable/2.x` branch:
> 
> We cherry-pick commits normally.

We cherry-pick to `2.1.x` and `2.2.x`. But `2.x` should point to one of those branches. We never do the release from that branch so it makes no sense for it to diverge from `2.2.x` (or whatever current minor version is for that major one).

PR #71, opened 2022-12-16
RFC 1018: Extending indexes

vpetrovykh:
Add a way to define parametrized abstract index. Also provide a way for users to define custom indexes extending existing ones while overriding some of the parameters.


vpetrovykh:

msullivan (approved):

PR #72, opened 2022-12-16
RFC 1019: Index fallback

vpetrovykh:
Add a way to define a fallback mechanism for indexes based on extensions being enabled or disabled.


vpetrovykh:

msullivan (approved):
Seems basically reasonable. I might end up wanting to argue about the syntax but no reason for that to block things

PR #74, opened 2023-01-04
RFC 1020: Multi choice control flow

aljazerzen:
One might say that this is not big enough change for a RFC, but hey let me have my fun.

I intentionally didn't push match in RFC title, because this is subject of the discussion.

### DIFF
--- a/text/1020-multi-choice-control-flow.rst
+++ b/text/1020-multi-choice-control-flow.rst
@@ -1,0 +1,76 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2023-01-04
+    Authors: Aljaž Mur Eržen <aljaz@edgedb.com>
+
+===========================================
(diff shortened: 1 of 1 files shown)


tailhook:
My major concern would be using `==` operator rather than full-blown pattern matching:
1. We mentioned somewhere that we probably want to implement algebraic types somehow at some point
2. Even without that it would be nice to match by string prefix, array patterns, and numeric ranges (Rust can do all these things, for example)

aljazerzen:
> rather than full-blown pattern matching

Yes, agree. The RFC does not even solve what the issue that requested it. My intention was to make something quick and extend it later.

Problem with that is that we should at least decide on syntax for pattern matching, otherwise we will probably run into backward incompatibility when adding it later.

Compiler team agrees that this RFC can wait and be extended to do pattern matching when algebraic types are a thing.

PR #75, opened 2023-01-18
RFC 1021: Mutation rewrites

msullivan:


text/1021-rewrites.rst
```diff
@@ -0,0 +1,210 @@
...
+
+Update a modification time field every time an object is modified, but
+allow a manual override of it also::
+
+  type Entry {
+      # ...
+      required property mtime -> datetime {
+          rewrite insert, update using (
+              datetime_of_statement()
+              if 'mtime' not in __fields__
```
fmoor, line 120: What about a special name for "this pointer's name". `__field__` or some thing. It seems like it would be easy to rename the property here but miss the string in the `if` which would break this rewrite.

text/1021-rewrites.rst
```diff
@@ -0,0 +1,210 @@
...
+<https://github.com/edgedb/edgedb/issues/4142#issuecomment-1386287450>_`.)
+
+Restrictions
+------------
+Each pointer can have one ``insert`` rewrite rule and one ``update``
+rewrite rule. They can be specified together if desired.
+
+Specifying a rewrite rule on a pointer is not allowed if an ancestor
+has already specified that kind of rewrite rule for the pointer.
+(Alternately: multiple rewrite rules are applied in MRO order?)
```
fmoor, line 78: what about the first rewrite rule found in MRO order is applied? This seems more intuitive to me.

If I think about this in terms of methods and object inheritance, which is probably the typical user's background then the first option sounds like "you can't define a method if an ancestor has already defined it" which doesn't seem helpful. The second option sounds like "every definition of a given method will be called in MRO order" which sounds even worse.

There are probably a lot of things I'm missing here though...

text/1021-rewrites.rst
fmoor: It seems like it would be possible to write a linter that complains about the the pointer name being used in a string in a rewrite when we have that kind of tooling in the future.

msullivan: Yeah, that's a good point.

The other thing I considered was just having a boolean variable that indicates whether the current pointer was specified (`__specified__`, maybe? I'm sure there is some decent name option.). I went with suggesting `__fields__` because it seemed more general but maybe that generality is not super helpful?

fmoor: It seems like having all the modified names would enable more use cases like only updating `friend_count` when `friends` is updated?

text/1021-rewrites.rst
elprans: I agree that overloading makes more sense here and is consistent with how you can overload a `default`.

text/1021-rewrites.rst
```diff
@@ -0,0 +1,210 @@
...
+            rewrite { insert | update } [, ... ] using (<expr>)
+        "}"
+    "}"
+
+We allow pointer definitions to be augmented with a ``rewrite`` rule.
+
+Expression interpretation
+-------------------------
+In ``<expr>``, ``__subject__`` (and the implicit path prefix) will be
+bound to the new value object being inserted or updated.
```
aljazerzen, line 55: Is referencing links, links properties or properties of link targets allowed?

text/1021-rewrites.rst
```diff
@@ -0,0 +1,210 @@
...
+
+In purely ``update`` rewrite rules, ``__old__`` will refer to the old
+version of the object.
```
aljazerzen, line 58: Is referencing links, links properties or properties of link targets allowed?

text/1021-rewrites.rst
```diff
@@ -0,0 +1,210 @@
...
+
+The special variable ``__fields__`` will be bound to all of the
+properties of ``<type-name>`` that were explicitly provided in the
+``insert`` or ``update``. This allows rewrite rules to distinguish
+between an explicitly provided empty set and no value being specified.
```
aljazerzen, line 63: Will `__fields__` contain values of other pointers that have already been rewritten?

A use-case:
```
type TodoList {
    multi link items -> Items;
    property is_empty -> bool {
        rewrite insert, update using (count(__subject__.items) = 0);
    }
    property status -> str {
        rewrite insert, update using (
            'done' if __subject__.is_empty else __subject__.status
        );
    }
}
```

If yes, in what order will they be evaluated and can they contain cycles?

The answer is obvious, the question is: do we want to support that?

tailhook:
Looking at @aljazerzen questions, I'm thinking, maybe it should be an extension of the `default` instead. Here are two examples of how it might look like:
```
property change_timestamp -> datetime {
    default := datetime_of_statement();
    use default on update;
}
```
Or:
```
property change_timestamp -> datetime {
    default (on insert, update) := datetime_of_statement();
}
```
Since it's a default `UPDATE .. SET { change_timestamp := $x }` obviously overrides it. And I don't think we need a syntax that explicitly overrides what user users. This is not column-based ACL.

It can also have other fields specified syntactically:
```
property cached_friend_count -> datetime {
    default := count(__subject__.friends);
    use default on update of friends;
    #                        ^^^^^^^ --- this field is tracked by a compiler
}
```
So renaming a link/property works in DDL and is a compiler error in SDL (unless property is fixed).

tailhook:
I also think that some examples, like this (including mine):
```
      multi link likes -> Like;
      required property cached_like_count -> int64 {
          rewrite insert, update using (count(.likes))
      }
```

Are better written as a **cached** computed property:
```
      multi link likes -> Like;
      cached property cached_like_count := count(.likes);
#     ^^^^^^ -- just a specifier on a computed property
```
In this case compiler can figure out itself that it depends on `.likes` and recompute when that link is changed.

This means that applicability of this `rewrite` is basically only on calling functions with side-effects. Which kinda an argument to make it just a simple extension of default like I've proposed above.

text/1021-rewrites.rst
elprans: Another option is to invent a function-like intrinsic, e.g. `specified(.mtime)`

elprans:
@tailhook one problem of extending `default` like this is that the semantics isn't clear. Generally, a "default" is used when a value is empty, whereas in the proposed design it would overwrite possibly non-empty values on `update`.

text/1021-rewrites.rst
msullivan: Hmmm. My intention was that all of the rewrites would be applied simultaneously, and all of them would refer to the original object

text/1021-rewrites.rst
vpetrovykh: I'm definitely +1 on overloading as the mechanism here. We opt for (you cannot overload this) only when we can't figure out how to do that reasonably and often it's a temporary restriction (because it's forward-compatible with overloading).

elprans:
Another capability that the RFC does yet show is the actual rewrite, i.e. the ability to alter specified incoming data.  One case would be providing data-format level compatibility, i.e. accepting some "old" format and automatically rewriting it into "new" format, thus supporting old API clients.  This is what a "default extension" approach wouldn't be able to support.

text/1021-rewrites.rst
msullivan: Yeah I think so

msullivan:
> Another capability that the RFC does yet show is the actual rewrite, i.e. the ability to alter specified incoming data. One case would be providing data-format level compatibility, i.e. accepting some "old" format and automatically rewriting it into "new" format, thus supporting old API clients. This is what a "default extension" approach wouldn't be able to support.

Oh yeah, this is a great point. I'll add an example along those lines.

msullivan:
The thing I don't love about the `cached` proposal is that it kind of promises more than it can deliver

text/1021-rewrites.rst
1st1: can we call it `__values__`?

aljazerzen: Another way would be to have __specified__ be a named tuple that contains booleans for each of the pointers:
```
... using ('had_status' if __specified__.status else 'did_not_have_status')
```

aljazerzen:
```
        rewrite on insert on update using (
            'done' if __subject__.is_empty else __subject__.status
        );
```

aljazerzen:
```
        rewrite (insert, update) using (
            'done' if __subject__.is_empty else __subject__.status
        );
```

aljazerzen:
```
        rewrite insert and update using (
            'done' if __subject__.is_empty else __subject__.status
        );
```

aljazerzen:
```
        rewrite insert update using (
            'done' if __subject__.is_empty else __subject__.status
        );
```

elprans (approved):
Looks good enough for me to get merged as the initial draft.

PR #76, opened 2023-01-23
1022: Typing free objects & simplifying SDL syntax. Initial draft

1st1:
This RFC proposes new type language syntax for declaring ad-hoc free object types as well as some modifications to our current SDL.

### DIFF
--- a/text/1022-freetypes.rst
+++ b/text/1022-freetypes.rst
@@ -113,14 +113,17 @@ Updates to SDL
 
 Summary:
 
-* Make the ``property`` and ``link`` keyword optional. The kind of a pointer
-  can be inferred automatically in all situations. Users will be encouraged to
-  explicitly specify the kind of a pointer for readability or enforcement
-  purposes where necessary.
+* Make the ``property`` and ``link`` keyword optional for all non-computed
+  properties and links. The kind of a pointer can be inferred automatically
+  in all situations. Users will be encouraged to explicitly specify the
+  kind of a pointer for readability or enforcement purposes where necessary.
 
-* Replace ``->`` symbol with ``:`` for type annotations of pointers and
-  functions return types. ``->`` will still be valid syntax, but we will
-  update our documentation and output of introspection commands to use ``:``.
+* Replace the ``->`` symbol with ``:`` for type annotations of pointers.
+  ``->`` will still be valid syntax, but we will update our documentation
+  and output of introspection commands to use ``:``.
+
+* Keep the ``->`` symbol for function return type annotation (in other words
+  we will not add support for using ``:`` in return type annotation).
 
 Motivation:
 
@@ -167,7 +170,7 @@ Example:
 |  }                                   |   }                                  |
 |                                      |                                      |
 |  type Show extending Content {       |   type Show extending Content {      |
-|    property num_seasons :=           |     num_seasons :=                   |
+|    property num_seasons :=           |     property num_seasons :=          |
 |      count(.<show[is Season]);       |       count(.<show[is Season]);      |
 |  }                                   |   }                                  |
 +--------------------------------------+--------------------------------------+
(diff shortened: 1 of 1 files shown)


text/1022-freetypes.rst
```diff
@@ -0,0 +1,213 @@
...
+        select # ...
+    )
+
+
+Updates to SDL
+==============
+
+Summary:
+
+* Make the ``property`` and ``link`` keyword optional. The kind of a pointer
```
msullivan, line 116: Unfortunately I don't think we can drop it in *all* places, since for computeds it will conflict with existing `SetField` commands. (For things like `default := "foo"`.)

I think we can drop it for non-computed uses, though I'm still fighting with the parser generator a bit

1st1: @msullivan Sully, since it looks like you made it work (with keeping "link" and "property" optional keywords) can you list all cases where link/property become optional and where they are needed to disambiguate?

1st1: (I'll update the RFC with that)

1st1: > Unfortunately I don't think we can drop it in all places, since for computeds it will conflict with existing SetField commands. (For things like default := "foo".)

That's only relevant for computable link properties though, right?

1st1: That said requiring "link" / "property" modifiers for all computeds is OK-ish compromise, I think.

text/1022-freetypes.rst
```diff
@@ -0,0 +1,213 @@
...
+
+In user-defined functions::
+
+    function compute(pokemon: {
+        required pokemon: str,
+        multi required locations: {
+            required point: tuple<float64, float64>,
+            street: str
+        }
+    }) -> int64 using (
```
tailhook, line 106: You're showing arrow here instead of `:` but declare:
> Replace ``->`` symbol with ``:`` for type annotations of pointers and functions return types.

That said, I would keep an arrow for function return types, because that improves readability. Or do we expect function signatures to be used in type casts anytime soon?

1st1: > Or do we expect function signatures to be used in type casts anytime soon?

Probably not. :)

> That said, I would keep an arrow for function return types, because that improves readability.

I'm not 100% against that. Rust is using arrows for return types, right? Python uses them too, but I think the main motivation for Python was the same one we have with whitespace-sensitive SDL -- Python can't use `:` for that.

I slightly prefer using `:` instead of `->` for function returns to minimize the surface of ASCII operators we have. If we go the `:` route all in we can just phase out `->` completely and pretty much forget that it exists in all our docs / tutorials / etc.

text/1022-freetypes.rst
msullivan: Yeah, the one case where we can't drop from a syntax perspective is `foo := ...`. This is kind of a bummer, but oh well.

Syntactically, we definitely can do `multi foo := ...`, but I think it might be worth *not* allowing that:
1. For consistency, it might be best to always just require `link`/`property` on computeds, instead of just allowing it to be dropped if there is a cardinality annotation
2. There are implementation reasons: for non-computed links, we should be able to resolve whether they are links or properties in `declarative.py`, when we convert SDL to DDL, and so wouldn't ever need to handle DDL where it is not known. For computeds it is trickier, since we would need to actually typecheck the expression. Handling DDL with unknown pointer kind is probably possible but probably kind of invasive.

text/1022-freetypes.rst
tailhook: I don't think we can ever get rid of `->` parsed as an operator for backwards compatibility reasons.

From Rust docs:
> ## [2.2 -> for function return type](https://doc.rust-lang.org/1.6.0/complement-design-faq.html#--for-function-return-type)
>
>This is to make the language easier to parse for humans, especially in the face of higher-order functions. fn foo<T>(f: fn(i32): i32, fn(T): U): U is not particularly easy to read.

Well, what is the chance that we will need first-class functions eventually?

elprans: > Well, what is the chance that we will need first-class functions eventually?

Almost 100% and soon (in 4.0 most likely)

I also think that `->` is a superior syntax for function types and that we should keep it.

PR #77, opened 2023-01-26
RFC 1023: Adding Splats Syntax

1st1:


text/1023-splats.rst
```diff
@@ -0,0 +1,257 @@
...
+      },
+
+      [is Hero].secret_identity,  # fetch `secret_identity` property
+                                  # for all `Person` objects that are also
+                                  # instances of `Hero`
+    }
+
+We propose to enhance the shape construct with the following syntax:
+
+* ``*``: extend the shape with all properties (not links!) defined on the type,
```
tailhook, line 63: I like everything except that this only fetches properties. I understand why it looks conceptually nice. But I think for REPL use case it's vastly superior to have links too:
```
> SELECT Person {*}
{
  name: "John",
  villains: [
    Villain { id: 4d1fcc65-8aa7-11ed-ad95-2325e2521606 },
  ]
}
```
This is how links (objects) are fetched by default, you see ids (and types) of the objects.

Otherwise it looks like you still have to run describe every time to find out what `*` didn't show.

elprans: I think I'm inclined to agree here.  The hesitation to include links in `*` is mostly dictated by performance worries, but then you are not really supposed to abuse `*` in application code, and for quick REPL queries and proofs-of-concept performance matters way less.

elprans: This implies that we invert the "no shadowing" rule, because if `*` includes links, elaboration of them in subshapes becomes impossible.

1st1: I still think that using `*` only for properties makes more sense for us for the following reasons:

* Users will inevitably use this in their application code (not just in REPL) and selecting links is expensive. Selecting all properties is typically a more common need than selecting all properties & all links. The performance would take a big hit if we select all links by default.

* We already have splats in JS (https://www.edgedb.com/docs/clients/js/querybuilder#select-objects, @colinhacks ❤️‍🩹) and the current API only expends `*` into properties.

* I have a feeling that when using splats in types (the second part of this RFC) it would be more valuable to unpack just properties.

That said I agree, that the REPL use case is important. So maybe we add `.**` syntax? @elprans suggests that potentially we can expand all links and all properties of all linked objects (1 level) to make it even more useful for exploration and messing around. Thoughts?

tailhook: Yes, double star is okay, I think, if it can also apply for standalone `User {**}`.

>  @elprans suggests that potentially we can expand all links and all properties of all linked objects (1 level) to make it even more useful for exploration and messing around. Thoughts?

Sounds plausible, but links are almost in all cases are cyclic. (i.e. computed `.<` is used too often, and we can't just ignore backlinks).

So I would like to have a way to have single level expansion and graph expansion as a separate operations, kinda:
1. `*` -- only properties
2. `*.*` -- properties and links
2. `*.**` -- links recursively, stopping when reaching the same type probably, i.e. `A -> B -> A` stops at the second `A`.

We can also probably invent some syntax for expanding fixed number of levels of links, like: `*.*.*` two level of links even if it's `User { friends: User }`.

But I'm not sure whether it's too much to learn.

tailhook:
Also maybe we should:
1. Have a session config option or compilation flag that enables splats? So we enable that in REPL but users have to make extra steps to use them in app code.
2. Should we prohibit splats in codegen? On one hand, it's convenient. But on the other hand they will eventually overfetch, so this is a bad habit to have. Also while property-only splats are maybe okay, link-including splats in codegen are almost surely a bad idea.

tailhook:
Also can we do something like this work:
```
WITH U = (SELECT User { first_name, last_name })
SELECT U {
    U.*,
    friends: {U.*},
}
```
I.e. some way to not repeat yourself in the query. Splats are a tool for that exactly, but...

text/1023-splats.rst
tailhook: While we're at it. It we a syntax that differ like `*` with properties only, and `**` or `*.*` that includes a link, we want show some hint:
```
> SELECT {*}
{User { name: "John" }}
# Asterisk `*` only shows properties, use `**` to also show links.
```
Which means we want to add some "annotation" to the `CommandDataDescription` when bare star is used.

1st1: I've updated the proposal with the `**` syntax.

I'm really not a fan of `*.*` and `*.**` variations. (FWIW @msullivan proposed something similar too). While they are certainly readable once you understand the mechanics, searching what they mean and how they work in the docs/web, as well as just quickly scanning through a query will be a pain for many users.

1st1:
@tailhook 

> Have a session config option or compilation flag that enables splats? So we enable that in REPL but users have to make extra steps to use them in app code.

I think it's a bad idea to gate any kind of syntax behind a flag. It will undoubtedly create a lot of confusion (why my query works here but not there). IMO this is a documentation issue.

> Prohibit splats in codegen? On one hand, it's convenient. But on the other hand they will eventually overfetch, so this is a bad habit to have. Also while property-only splats are maybe okay, link-including splats in codegen are almost surely a bad idea.

Theoretically we could include a flag to our type descriptor that it included a splat...But I'm not sure that we should do this in 3.0. Maybe a request for this should come from the community.

> WITH U = (SELECT User { first_name, last_name })
SELECT U {
    U.*,
    friends: {U.*},
}

No, this wouldn't work because the `U` alias doesn't *define a new type*. The type of `U` is still `User` with all its properties. You depend on that property of EdgeQL in your query, btw, selecting the `friends` link via the `U` alias.

elprans:
One thing that still needs an explicit mention/discussion is this pattern:

```
WITH
   U := User { last_name := str_upper(.last_name) }  
SELECT (
   U { * },          # will select the `last_name` computed
   
   U { User.* },     # should this select the non-computed version of `last_name`?
                     # or should it be equivalent to U { U.* }?
```

I think making `Child { Parent.* }` equivalent to "select all properties found in `Parent`, but as they are defined in `Child`" is safer, but there could be a use case of "escaping the computed"?

1st1:
> should this select the non-computed version of `last_name`?

IMO no, it should include the computed one. All `User.*` does is expanding `U { User.* }` to `U { first_name, last_name, email, etc }`.

text/1023-splats.rst
```diff
@@ -0,0 +1,340 @@
...
+     function validate(data: {
+       required Hero.*  # will expand into:
+                        #   {
+                        #     required id: uuid,
+                        #     required name: str,
+                        #     required secret_identity: str
+                        #   }
+     }) -> bool using (...)
+
+When we add a ``never`` type eventually (to pave the path to implementing
```
tailhook, line 254: Oh, this is not how `never` works usually.

I.e. a thing that contains `never` can never exists. Let me explain it on tuples first.

One way to think of it intuitively is: if you have a function that returns never:
```
function raise_exception() -> never {..}
```
And use that function:
```
SELECT ("test", raise_exception())
```
This tuple is never created, so compiler is free to optimize it out to `SELECT raise_exception()` (this optimization is a bit more complex that it looks like but I'll omit details for brevity).

Now there is a dilemma:
1. Objects don't behave like tuples in our language (with regard to cross-product semantics)
2. But still if you write `SELECT { name := "test", other := raise_exception() }` the free object `{ name: str, other: never }` will never be created and exception raised, so the return type of this `SELECT`-expression is equivalent to `never`.

So unless I'm wrong with semantics of never/exceptions in (2) we should not use that syntax to cherry-pick fields to exclude.

tailhook:
> I think it's a bad idea to gate any kind of syntax behind a flag. It will undoubtedly create a lot of confusion (why my query works here but not there). IMO this is a documentation issue.

We always disagree on how much we can rely on the documentation. But I don't think there will be "lot of confusion", because it's super-simple to write an explanation in the error message:
```
>>> conn.query("SELECT User {*}");
CannotUseAsteriskError: usage of asterisks in the queries is disabled
| SELECT User {*}
               ^ --- please expand to specific attribute names
hint: Properties of the User: id, first_name, last_name
hint: Links of the User: friends, profile, publications
help: This feature is developed for exploration in shell and exploration tools. But it's
help: better to enumerate specific fields in application code when using client library.
help: Alternatively enable `allow_asterisks` option.
more info: http://edgedb.com/p/asterisk
```
You can even be lazy enough to put star, see exception and copy-paste all the properties from the exception message.

> Theoretically we could include a flag to our type descriptor that it included a splat...But I'm not sure that we should do this in 3.0. Maybe a request for this should come from the community.

I think annotation would be good enough. Also as I've already said, we should have an annotation for `*` to suggest user trying `**` in REPL.

> No, this wouldn't work because the U alias doesn't define a new type. The type of U is still User with all its properties. You depend on that property of EdgeQL in your query, btw, selecting the friends link via the U alias.

Sure. My point is that do-not-repeat-yourself is another super-important function of splats that we have to add eventually. If we don't want that syntax maybe we can have a special constructor for that:
```
WITH FRAGMENT Common := {first_name, last_name}
SELECT User { Common.*, friends: { Common.* }}
```

One interesting way of how bare "fragments" (the terminology stolen from GraphQL) can be used is kinda "reverse codegen":
```rust
#[derive(Queryable)]
struct BaseUser {
    first_name: String,
    last_name: String,
}

#[derive(Queryable)]
struct User {
    #[edgedb(flatten)]
    common: BaseUser,
    friends: Vec<BaseUser>,
}

let users: Vec<User> = query!(conn, "
    WITH FRAGMENT BaseUser := <some subst syntax to get fields from `struct BaseUser`>
    SELECT User { BaseUser.*, friends: { BaseUser.* } }
").await?
```

*(There is a number of reasons of why it's more convenient to define structure in Rust rather then using codegen from EdgeDB query to Rust code)*

text/1023-splats.rst
1st1: > This tuple is never created, so compiler is free to optimize it out to SELECT raise_exception() (this optimization is a bit more complex that it looks like but I'll omit details for brevity).

Hm, that's an interesting observation. FWIW I copied the TypeScript behavior w.r.t. `never` when writing this up. In TypeScript you can use `never` like this specifically in the *type* subset of the language. In the *type* language, `never` behaves differently from *runtime*, only behaving as a hint during the computation of the whole type expression.

That said, I think you're right, in our semantics it would be a bad idea to use `never` the way I suggested in this proposal. Good catch, Paul.

Perhaps a variation of this syntax could work:

```
function foo(param: {
  User.*,  # splat all fields from the User type
  -id,     # exclude the `id` field
}) 
```

1st1:
> I think annotation would be good enough. Also as I've already said, we should have an annotation for * to suggest user trying ** in REPL.

Yeah, with annotations clients will be able to implement whatever API they want, including having a client-level setting to ban splats.

> Sure. My point is that do-not-repeat-yourself is another super-important function of splats that we have to add eventually. If we don't want that syntax maybe we can have a special constructor for that:

> WITH FRAGMENT Common := {first_name, last_name}
SELECT User { Common.*, friends: { Common.* }}

Maybe we don't need that. We do want to add support of defining type and type aliases in the `with` clause. With that and with free objects you'd be able to do this:


```
with
  shape as type {
    first_name: str,
    last_name: str
  }
select {
}
```

tailhook:
> Maybe we don't need that. We do want to add support of defining type and type aliases in the with clause.
> ```
> with
>  shape as type {
>    first_name: str,
>    last_name: str
>  }

The issue I see here is that you have to restate the types of the fields. This probably defeats the purpose of not repeating yourself, because this will require syncing types here to types in the schema.


> Lastly, maybe we should prohibit using splats on anything but types.

This has a problem with computeds:
```
WITH U := (SELECT User { full_name := .first_name ++ " " ++ .last_name })

SELECT {
  users_with_full_name := (SELECT User { U.* }),
  users_without_full_name := (SELECT User { User.* }),
}
```

elprans:
> The issue I see here is that you have to restate the types of the fields

We may consider allowing incomplete free types, e.g. with `anytype`:

```
with
 fragment as type {
   foo: anytype,
   bar: anytype,     
 }
```

and, possibly even allow eliding `anytype`:

```
with
  fragment as type {
    foo,
    bar,
  }
```

> This has a problem with computeds:

No problem, as suggested above, `(typeof U).*` should still work.  The reason here is that `.*` is a _type operator_, not a _value operator_ (in the same vein as the infix `IS` or `INSERT` are).

msullivan:
Does `**` include `__type__` or should we explicitly exempt that?
It *is* a link, but also it has 15 properties and the user *maybe* care about name (but probably not, since name is already injected in a much more convenient way).
(Properties of `schema::ObjectType`: `["final", "expr", "is_from_alias", "builtin", "is_compound_type", "id", "inherited_fields", "from_alias", "is_abstract", "internal", "computed_fields", "is_final", "abstract", "name", "compound_type"]`)

tailhook:
> No problem, as suggested above, `(typeof U).*` should still work. The reason here is that `.*` is a _type operator_, not a _value operator_ (in the same vein as the infix `IS` or `INSERT` are).

Currently you can do `something is U`, it always returns `false`. So it looks like a bug, right? I mean either `U` is *also* a type, and you can do `is U` or, `U` is not a type and in this case `is U` is a compilation error.

tailhook:
> Does `**` include `__type__` or should we explicitly exempt that?

I think exemption is good. Probably `SELECT (introspect typeof X).*` is a way to find out details about the type.

elprans:
> Currently you can do `something is U`, it always returns `false`. So it looks like a bug, right? I mean either `U` is _also_ a type, and you can do `is U` or, `U` is not a type and in this case `is U` is a compilation error.

This seems like a bug to me.  @msullivan?

1st1:
> Does ** include __type__ or should we explicitly exempt that?

Definitely exempt.


>>    Currently you can do something is U, it always returns false. So it looks like a bug, right? I mean either U is also a type, and you can do is U or, U is not a type and in this case is U is a compilation error.

> This seems like a bug to me. @msullivan?

Yeah, sounds like this should have been a compile-time erorr

text/1023-splats.rst
tjjfv: A suggestion for the syntax:
- `*` for properties
- `*>` for properties and links
- chaining for specific depth, e.g. `*>*`, `*>*>`,`*>*>*`
- `***` for recursive

*(I'm new to EdgeQL, and I recognize this may not be good or viable for a variety of reasons.)*

elprans: We discussed the possibility of adding syntax for deeper (arbitrary) expansions, but decided against it (at least for now), because:

- the syntax is ends up a bit arcane looking
- the real need for this is not entirely clear

@1st1, let's add a note to the RFC that this has been considered.

elprans:
@1st1, @msullivan, one thing we forgot about here are the usual middle children: link properties.  I think `**` should include them, but this deserves a mention in the RFC.

elprans:
Another missing bit is the combination of polymorphic intersection and explicit type source, e.g. `Foo[IS Bar].*`

tailhook:
> Another missing bit is the combination of polymorphic intersection and explicit type source, e.g. `Foo[IS Bar].*`

Oh, there is actually two questions here:
1. Whether `Foo[IS Bar].*` supported? I would say it's unlikely useful in ad-hoc repl queries. And we don't want asterisks in production queries.
2. Whether `Foo {*}` expands to `Foo { Foo.*, [IS Bar]Bar.*, [IS Baz]Baz.* }`, which sounds as something hugely useful and nice.

PR #79, opened 2023-02-03
Update the RFC 1015.

vpetrovykh:
Change FTSResult to a named tuple.
Move `language` from an annotation to a parameter. Rename `boost` to `weights` and make it a parameter to the search functions.
Remove the search functions that accept Objects, instead use a variadic `doc` parameter to accept one or more properties.

### DIFF
--- a/text/1015-full-text-search.rst
+++ b/text/1015-full-text-search.rst
@@ -58,11 +58,11 @@ Specification
 =============
 
 There's a lot of functionality involved in supporting FTS, but it's a fairly
-niche thing, so it makes sense to create a ``module fts`` for all of the FTS
-tools.
+niche thing, so it makes sense to create a ``module std::fts`` for all of the
+FTS tools.
 
 In EdgeDB we should avoid having an explicit "document" scalar like
-``tsvector``. Instead we can rely on various parameters, annotations and
+``tsvector``. Instead we can rely on various parameters and
 indexes to specify how the plain text ``str`` should be parsed into tokens.
 This will allow us to keep the queries involving FTS mostly implementation
 agnostic. Some FTS backends may have functions specific to them, but the goal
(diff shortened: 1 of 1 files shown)


text/1015-full-text-search.rst
```diff
@@ -283,50 +258,38 @@ minimum and maximum prefix values.
...
+specific backend.
 
-    required property content -> str;
-    index on (.content) using fts;
-  }
+The weight value is basically like a relevance score multiplier or a hint of
+how much search matches in a certain field should be prioritized compared to
+other fields. This concept is necessarily fuzzy as it has to account for
+various possible backends. This value can be passed as an optional ``weights``
+parameter to FTS search functions.
```
aljazerzen, line 284: Didn't we settle on `[0, 1]` for the range of weights?

text/1015-full-text-search.rst
```diff
@@ -73,59 +73,57 @@ queries rather than introducing a new scalar type. This string is expected to
...
+used as the default for this parameter. Alternatively, the default language
+may be set via config values. If either of these options is implemented, the
+``language`` parameter may be omitted.
+
+The ranked results need to have richer information and thus named tuples are
+used::
+
+  function std::fts::match(
+      query: str,
+      variadic doc: optional str,
```
msullivan, line 107: (`variadic` + `optional` is definitely workable for a built-in SQL defined function like `fts::match` but seeing this made me realize that we allow it in edgeql and it does not behave sensibly (https://github.com/edgedb/edgedb/issues/4982))

msullivan (approved):

text/1015-full-text-search.rst
```diff
@@ -73,59 +84,192 @@ queries rather than introducing a new scalar type. This string is expected to
...
+Search Functions
+----------------
+
+An important feature of the search function is that it actually performs the
+job typically done by the ``FILTER`` clause. The function is supposed to
+filter a bunch of objects based on the FTS query and indexed fields. In
+addition to filtering, the results may be annotated with relevance (ranked)
+and potentially with highlighted matches::
+
+  function std::fts::search(
```
msullivan, line 234: Let's turn `docs: set of anyobject,` into just `doc: anyobject`, and change the return type from `set of` to `optional`.

I think it should produce essentially the same behavior in most use cases, but is cleaner.

text/1015-full-text-search.rst
```diff
@@ -73,59 +84,192 @@ queries rather than introducing a new scalar type. This string is expected to
...
+      docs: set of anyobject,
+      query: str,
```
msullivan, line 236: I believe we decided we needed a `language` param also

text/1015-full-text-search.rst
```diff
@@ -73,59 +84,192 @@ queries rather than introducing a new scalar type. This string is expected to
...
+Postgres).
+
+In principle the group names can be generalized to be any ``str`` and the
+lexicographical order can then be used to determine the relative group order
+(for purpose of picking group weights). This would allow using Postgres-style
+groups 'A', 'B', 'C', 'D' or give groups more descriptive names like '1 - most
+relevant', '2 - regular', '3 - marginal'.
+
+
+Inheritance
```
msullivan, line 192: We want to make conflicting indexes inherited from different base types into an error

text/1015-full-text-search.rst
```diff
@@ -73,59 +84,192 @@ queries rather than introducing a new scalar type. This string is expected to
...
+      body: str;
+      internal: str;
 
-  function fts::match(
-    doc: str,
-    query: str,
-    rank_opts: optional str = 'default',
-    highlight_opts: optional str = {}
-  ) -> FTSResult
+      index fts::textsearch(language := 'english')
```
msullivan, line 165: We probably want `fts::weight` and friends to return some opaque internal type. We'll disallow having it appear in non-index queries

msullivan (commented):
Feedback from our call today

text/1015-full-text-search.rst
```diff
@@ -58,74 +58,266 @@ Specification
...
+that affect how the string is interpreted without having to change all the
+signatures.
+
+This processed document type and a couple of functions that use it are::
+
+  type fts::DocVector {
+    # This is hidden implementation, so it's more of a concept
+    # than an actual object type. Under the hood it's likely
+    # to be a record, though.
+    required docuemnt: str;
```
1st1, line 93:
```suggestion
    required document: str;
```

text/1015-full-text-search.rst
```diff
@@ -58,74 +58,266 @@ Specification
...
+opaque and not valid as a result of a user query. Instead it can be used in
+index expressions and returned by functions that can only be used in index
+expressions. The type being opaque makes it possible to add more FTS functions
+that affect how the string is interpreted without having to change all the
+signatures.
+
+This processed document type and a couple of functions that use it are::
+
+  type fts::DocVector {
+    # This is hidden implementation, so it's more of a concept
```
1st1, line 90: What does it mean "hidden"? Will it be opaque but still available? If it's opaque, how will you deal with it lacking introspection?

1st1: I guess "index" functions won't be introspectable and that kinda solves this issue?

text/1015-full-text-search.rst
vpetrovykh: I've put back the `doc: anyobject`.

text/1015-full-text-search.rst
vpetrovykh: Added parameter

text/1015-full-text-search.rst
vpetrovykh: Added a note about that in inheritance.

text/1015-full-text-search.rst
vpetrovykh: I've added more explanation regarding the type. Any function works with this type should also be excluded from introspection.

text/1015-full-text-search.rst
vpetrovykh: Fixed this an a bunch of other typos.

text/1015-full-text-search.rst
```diff
@@ -58,74 +58,272 @@ Specification
...
+details of which properties should be searched and other FTS parameters will
+be provided by the FTS index on the specified type. Searching an unindexed
+type should simply produce no matches (resulting in an ``{}``).
+
+The ``language`` parameter indicated which language the query is using and
+therefore allows to target only the relevant documents if there are multiple
+languages.
+
+If ranking was requested (which is the default behavior), then the value will
+be written in ``rank``, otherwise the value ``0`` is used.
```
aljazerzen, line 306: Rank will always be requested, right?

It might not be used in the result of the query, in which case the compiler can make sure it is not even computed, but it will never default to 0.

PR #83, opened 2023-08-09
RFC 1024: Index ordering

vpetrovykh:
Provide a way to specify ordering for the index expressions. Also make the support of indexes on multiple fields more consistent (stop using tuples for that).

### DIFF
--- a/text/1024-index-ordering.rst
+++ b/text/1024-index-ordering.rst
@@ -1,0 +1,95 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2023-08-09
+    Authors: Victor Petrovykh <victor@edgedb.com>
+
+========================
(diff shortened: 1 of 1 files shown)


text/1024-index-ordering.rst
```diff
@@ -0,0 +1,95 @@
...
+can be translated into::
+
+  CREATE INDEX foo_index ON "Foo" (x DESC NULLS LAST, y ASC);
+
+
+Backwards Compatibility
+=======================
+
+This change is *syntactically* backwards compatible since all of the extra
+specification for the index are optional. However, sematically, we should not
```
elprans, line 78: I don't think purity is worth the cost of breaking compatibility and requiring a migration. IMO we should continue to special-case tuples.

PR #84, opened 2023-11-01
Update the RFC 1015

vpetrovykh:
Bring the RFC document and implementation into sync. We've changed some names of various `fts` parts and added new enums.

### DIFF
--- a/text/1015-full-text-search.rst
+++ b/text/1015-full-text-search.rst
@@ -84,30 +84,25 @@ expressions. The type being opaque makes it possible to add more FTS functions
 that affect how the string is interpreted without having to change all the
 signatures.
 
-This processed document type and a couple of functions that use it are::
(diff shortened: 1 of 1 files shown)


text/1015-full-text-search.rst
```diff
@@ -122,6 +117,18 @@ Algolia has an ``indexLanguages`` setting for its indexes.
 
 Typesense does not appear to have any language-specific index settings.
 
+We define some built-in enums to denote available languages. The most general
+is ``fts::Language`` and it is listing languages that are supported across
+backends. There are also backend-specific enumssuch as ``fts::PGLanguage`` and
```
aljazerzen, line 122:
```suggestion
backends. There are also backend-specific enums such as ``fts::PGLanguage`` and
```

aljazerzen (approved):

PR #85, opened 2024-01-16
1025: Database branches draft

vpetrovykh:


text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+   branch to it. Only the schema is copied, the data is still empty and needs
+   to be populated separately. This parallels source control branches that
+   diverge in terms of the schema. It is recommended to setup a new database
+   branch whenever working on a code branch that alters the schema in some
+   way. This will help reduce friction when switching between code branches by
+   having database states that correspond to the schema and avoiding
+   accidental mismatches in migrations.
+
+   If the ``from <oldbranch>`` part is omitted, the new branch is created
+   using the branch from the current connection.
```
msullivan, line 67: I think that @1st1 specifically didn't want this, and wanted to force being more explicit

text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+When rebasing one branch on top of another we can use introspection to compare
+the respective migration histories and find the point where they diverge.
+Afterwards we can try to apply a batch of new migration to the existing branch
+and if there are no issues perform a "fast-forward" rebase.
+
+In order to minimize the hassle of rebasing the git branches corresponding to
+database branches we need to give migration files names that are distinct in
+these branches so that git does not attempt to merge the file contents. This
+may be either directly the branch name appended to the numeric name or a
+(short) hash derived from the branch name.
```
msullivan, line 136: We can also potentially avoid needing to have distinct migration file names by having a little bit of VCS integration. Not sure if we want to, though?

1st1: I think VCS integration should be strictly optional.

> This may be either directly the branch name appended to the numeric name or a (short) hash derived from the branch name.

I think @elprans proposed hashing the message part of the migration?

text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+
+There are several new DDL commands introduced by this proposal:
+
+1) ``create empty branch <newbranch>``
+
+   The most basic command creates a new branch with an empty schema
+   (exactly what ``create database`` command currently does).
+
+2) ``create branch <newbranch> [ from <oldbranch> ]``
+
```
haikyuu, line 56: I guess this would be translated into `create branch <nb> from <ob>@state`.

In development, one regular worfklow is to checkout to an older commit, create a new branch and start from there. I guess in that case, it will pick the branch migration state from the schema file? or how would that work?

aljazerzen: Database branches do have their "current state", so it would be using that state.

A checkout of arbitrary state would be possible to implement, but would be slow, as we would have to create an empty branch and apply all needed migrations one-by-one.

There would be no way of keeping any data contents.

text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+   Removes an existing branch from the instance.
+
+
+Project and CLI Integration
+---------------------------
+
+Currently the project configuration file ``edgedb.toml`` does not have to
+specify the default *database* it connects to. With this proposal specifying
+the branch is mandatory.
```
aljazerzen, line 79: This might want to expand the reasoning for this:

> Intended usage is to create a new database branch when work on a new VCS branch is started. When switching VCS branches, `edgedb.toml` would be changed as well, changing the used database branch.

text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+Instead we want to enhance the developer experience while using multiple
+source control branches. These multiple databases could map well onto
+corresponding source branches, representing slightly different version of the
+data and schema. Thus there is a good reason to rename the concept of
+*database* to *branch* and provide additional branch-handling tools. With this
+change it is also now no longer necessary to have a semantic distinction
+between the terms *instance* and *database* which may be natural for some
+users to think of as interchangeable.
+
+
```
aljazerzen, line 44: We should write down that "even though similar to VCS, branches would be gain the ability to be reset to a previous state".

text/1025-branches.rst
```diff
@@ -0,0 +1,181 @@
...
+the respective migration histories and find the point where they diverge.
+Afterwards we can try to apply a batch of new migration to the existing branch
+and if there are no issues perform a "fast-forward" rebase.
+
+In order to minimize the hassle of rebasing the git branches corresponding to
+database branches we need to give migration files names that are distinct in
+these branches so that git does not attempt to merge the file contents. This
+may be either directly the branch name appended to the numeric name or a
+(short) hash derived from the branch name.
+
```
aljazerzen, line 137: This is how I understand this process:

Main (VCS and database) branch contains migrations M1, M2 and M3.
They are written in files named 00001.edgeql, 00002.edgeql, 00003.edgeql, respectively.
These filenames do not contain branch name, as that name is equal to `main`.

Now we start a new (VCS and database) branch feat-a for feature A and
create migrations A1, A2 and A3,
in files feat-a-00004.edgeql, feat-a-00005.edgeql and feat-a-00006.edgeql.
These files might have been added in a single or in multiple VCS commits.

In parallel, we continue development on main branch with migrations M4 and M5, in files
00004.edgeql and 00005.edgeql.

Now we want to merge (rebase) feature A into main.
Normal VCS process would start on branch main and merge files changes from feature A.
Because migration files contain branch names, this will not merge the two 00004.edgeql files into one.

After this step, the file system would contain files 00001.edgeql, ..., 00005.edgeql,
as well as feat-a-00004.edgeql, feat-a-00005.edgeql and feat-a-00006.edgeql.

Because files feat-a-00004.edgeql and 00004.edgeql both reference M3 as their parent, we need
to rebase all of feat-a files onto M5.
For this, we'd introduce `edgedb branch rebase feat-a`, which would:
- read all migration files,
- identify all migrations from feat-a, by filenames,
- verify that all remaining migrations are linear and have been applied to the database,
- recreate feat-a migrations, one by one,
- write them to 00006.edgeql, 00007.edgeql and 00008.edgeql,
- delete original feat-a files.

text/1025-branches.rst
aljazerzen: What's the problem with branch names in migration files?

Just by looking at the file system, it is hard to determine which migration is the "last" one and should match with the state of the database?

msullivan: The issue is that when we are doing an `edgedb migration rebase` (or whatever), it will probably be while trying to resolve a `git rebase`/`git merge`. Since both branches will have migrations not in the other in the interesting case, we will have `00010.edgeql` (or whatever) filled with merge conflict markers, and so we can't necessarily reconstruct anything from it.

That said, I don't totally remember why exactly we decided we needed information from the file system in order to do an `edgedb migration rebase`. We can read the migration histories out of the database branches themselves, possibly, and find the common ancestor that way?

I have a nagging feeling that we found some fatal flaw in that plan, though, but I can't remember it, and I think I left all of my notes from the offsite in my hotel room...

text/1025-branches.rst
msullivan: The RFC hasn't been updated yet, but we decided on the call yesterday that storing the current branch in `edgedb.toml` is probably a mistake, because it will be too easy to accidentally merge the config for a feature branch back into main.

What we were thinking of doing instead in storing the db branch name in `.edgedb.local.toml` or some such, which should be in `.gitignore`. The CLI could offer to put it in `.gitignore` for you, and it can also offer to install a `post-checkout` hook script that automatically updates `.edgedb.local.toml` when changing branches.

text/1025-branches.rst
```diff
@@ -0,0 +1,270 @@
...
+---------------------
+
+We should eventually have a way to simplify working with git branches and
+keeping the database branches synchronized. Merging or rebasing of database
+branches can also benefit from git integration in order to correctly identify
+the migration history and create reasonable migration files. We can also
+possibly apply rebasing logic in smaller steps as we could access the
+intermediate schema states from the git commit history.
+
+We might add a ``using copy_data`` clause to ``create branch <newbranch> from
```
1st1, line 178: We need to bikeshed on syntax here, `copy_data` is 😱

text/1025-branches.rst
```diff
@@ -0,0 +1,270 @@
...
+
+   Removes an existing branch from the instance.
+
+
+Project and CLI Integration
+---------------------------
+
+Currently the project configuration file ``edgedb.toml`` does not have to
+specify the default *database* it connects to. We  can add a
+``edgedb.local.toml`` config file that can be used to supplement and override
```
msullivan, line 82: `.edgedb.local.toml` (leading dot)

text/1025-branches.rst
aljazerzen: I've mistyped and edited now.

I'm saying that because our branches are similar to VCS, one would expect them to also support a `git reset --hard HEAD~1`.

elprans (approved):

PR #86, opened 2024-01-23
RFC: 1026 function overloading

aljazerzen:


text/1026-function-overload.rst
```diff
@@ -0,0 +1,177 @@
...
+   create function foo(a: str) -> str using ("matched str");
+
+   select foo("hello"); # {"matched str"} select foo(12); # {"matched
+   anyscalar"}
+
+Note that in user-defined functions, it is not allowed to use
+pseudo-types or overload functions that take object types. Overloading
+of functions that take scalar types does not work correctly. This means
+that this overloading behavior cannot used outside of EdgeDB-defined
+functions.
```
aljazerzen, line 93: T

text/1026-function-overload.rst
```diff
@@ -0,0 +1,177 @@
...
+
+   create function foo(a: anyscalar) -> str using ("matched anyscalar");
+   create function foo(a: str) -> str using ("matched str");
+
+   select foo("hello"); # {"matched str"} select foo(12); # {"matched
+   anyscalar"}
+
+Note that in user-defined functions, it is not allowed to use
+pseudo-types or overload functions that take object types. Overloading
+of functions that take scalar types does not work correctly. This means
```
aljazerzen, line 91: This is what happens with scalars currently:

```
      create scalar type A extending str;
      create scalar type B extending A;
   
      create function foo(a: str) -> str using ("matched str");
      create function foo(a: A) -> str using ("matched A");
   
      select foo(<str>"hello"); # error: please disambiguate between the two function
                                # I would say this is incorrect, it should "match str"

      select foo(<A>"hello");   # matched A
                                # correct

      select foo(<B>"hello");   # internal server error: function edgedbpub."..." is not unique
                                # obviously bad
```

PR #90, opened 2024-03-06
Remove "as the name implies"

raddevon:
I don't think the name does imply this. Maybe the name was changed after this was written?


msullivan:
It used to be `local`, but also there is a pending PR that just totally removes it, since we are dropping it entirely.

PR #91, opened 2024-03-07
Update the switch command (RFC 1025)

quinchs:


1st1:
Can you rebase thi on top of the latest master. And update the PR to say that `branch create` does not switch the active branch to then newly created one.

PR #95, opened 2024-08-14
wip RFC 1026: A networking module

scotttrinh:


text/1026-net-module.rst
```diff
@@ -0,0 +1,208 @@
...
+The module will provide the following functions:
+
+1. HTTP Request Function:
+
+.. code-block:: edgeql
+
+   scalar type std::net::HttpMethod extending std::enums<GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS>;
+
+   function std::net::http_request(
+       method: std::net::HttpMethod,
```
1st1, line 55: I'd consider making method 'named only' and defaulting to 'get'.

text/1026-net-module.rst
```diff
@@ -0,0 +1,208 @@
...
+       url: str,
+       body: optional str,
+       named only headers: optional json
```
1st1, line 58: Why not `array<str>`? Easier to work with and type IMO.

scotttrinh: Wait... What would that look like?

```edgeql
   with
       payload := '{"key": "value"}',
       response := (
           select std::net::http_request(
               'POST',
               'https://api.example.com/webhook',
               body := payload,
               headers := ["Content-Type: application/json"],
           )
       )
   select response {
       id,
       state,
       created_at,
   };
```

—?

text/1026-net-module.rst
```diff
@@ -0,0 +1,208 @@
...
+=======================
+
+This RFC introduces new functionality and does not affect existing
+features. There are no backwards compatibility issues.
+
+Security Implications
+======================
+
+1. URL Validation: Implement proper URL validation to prevent malicious
+   URL inputs.
```
1st1, line 184: I'd rephrase -- "URL validation" implies that we somehow verify the correctness of a URL right before the request is made. What we should do instead is having a std::URL type that would implement correct construction and quoting/concatenation.

text/1026-net-module.rst
```diff
@@ -0,0 +1,208 @@
...
+2. Rate Limiting: Consider implementing rate limiting to prevent abuse
+   of the networking functions.
+3. TLS: Ensure that all network communications use secure protocols
+   (HTTPS, SMTPS) by default.
```
1st1, line 188: What if you explicitly specify `http://` in your URL? Will we assume https?

text/1026-net-module.rst
```diff
@@ -0,0 +1,208 @@
...
+
+Rejected Alternative Ideas
+--------------------------
+
+1. Using pg_net: While pg_net provides similar functionality, it was
+   decided to implement our own solution for better control and
+   integration with EdgeDB. This allows end users to more easily scale
+   sending by scaling the EdgeDB server rather than scaling PostgreSQL.
```
1st1, line 196: I'd also mention the security aspect -- handling HTTP in EdgeDB better isolates the actual database layer from vulnerabilities in HTTP libraries. That said, I'm not 100% sure this is a very sound argument :)

text/1026-net-module.rst
scotttrinh: Yeah, maybe "ensure" is not the right word. Utilize, but maybe this is a non-sensical point: the protocol itself ensures the security of the protocol, not really part of our purview.

text/1026-net-module.rst
1st1: yeah, exactly that.

scotttrinh: That feels slightly less helpful since `["hello!"]` is also an `array<str>`, but I suppose `json` is even worse in that regard. Maybe `array<tuple<key: str; value: str>>` is the safest at the expense of some additional syntax?

```edgeql
               headers := [("Content-Type", "application/json")],
```

wdyt?

1st1: tuple works!

1st1: nitpicking: also the actual type should probably be `array<tuple<name: str; value: str>>`

scotttrinh: Yeah, I really like this change, great suggestion.

text/1026-net-module.rst
```diff
@@ -0,0 +1,200 @@
...
+
+The module will provide the following functions:
+
+1. HTTP Request Function:
+
+.. code-block:: edgeql
+
+   scalar type std::net::HttpMethod extending std::enums<GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS>;
+
+   function std::net::http_request(
```
scotttrinh, line 54: @1st1 

What do you think of each protocol having a separate submodule here to organize the related types and functions? So `net::http::request` and `net::http::smtp`, etc?

text/1026-net-module.rst
```diff
@@ -0,0 +1,200 @@
...
+Functions
+---------
+
+The module will provide the following functions:
+
+1. HTTP Request Function:
+
+.. code-block:: edgeql
+
+   scalar type std::net::HttpMethod extending std::enums<GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS>;
```
scotttrinh, line 52: `GET` and `DELETE` are reserved words. Does this make it more awkward to use due to needing to quote it, or is it fine since `GET` is the default and `DELETE` is relatively less common to use?

text/1026-net-module.rst
```diff
@@ -0,0 +1,200 @@
...
+
+ResponseState enum
+------------------
+
+Since network requests are asynchronous, the response state will be
+represented by an enum:
+
+.. code-block:: edgeql
+
+   scalar type net::ResponseState extending std::enums<Pending, InProgress, Complete, Failed>;
```
fantix, line 82: There could be some difference between the status of a response and the state of the requesting task, and such a difference may affect how we retry the request.

A task could have such states:

* `Pending`
* `InProgress`
* `Complete`
* `Failed`
* `Error`

I added an `Error` state for non-retryable failures, like a bad request, too many retries, timed out in receiving response (server may have accepted the request without us knowing). In this case, we can only retry on `Failed` state (transitioning to `InProgress`), or certain `Complete` cases depending on the response status below.

Meanwhile, a response may have different statuses (based on the HTTP status):

* `Success` (200-ish)
* `ClientError` (400-ish)
* `ServerError` (500-ish)

By default, I think we shouldn't retry on client errors, but it may be safe to always retry on server errors. Still, this would be very much domain-specific (the remote may not follow the HTTP semantics as we assumed), so ultimately, we may have to let users decide when to retry. However, perhaps we don't need an additional `HttpStatus` now (there is already the `status` field), especially for the `try_request` MVP. I just feel wrong mixing this response status with the task state. Like the task state can be shared between `http`, `smtp` or friends, but the response status is protocol-specific.

text/1026-net-module.rst
```diff
@@ -0,0 +1,200 @@
...
+.. code-block:: edgeql
+
+   scalar type net::HttpMethod extending std::enums<`GET`, POST, PUT, `DELETE`, PATCH, HEAD, OPTIONS>;
+
+   function net::http_request(
+       url: str,
+       named only body: optional str,
+       named only method: net::HttpMethod = net::HttpMethod::GET,
+       named only headers: optional array<tuple<name: str, value: str>>
+   ) -> net::HttpResponse;
```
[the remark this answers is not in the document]
scotttrinh: > the task object links to both the request (for auditing purposes) and response object (more on the different states later)

Yeah, that's a good call. I would naturally reach for some kind of generic type in normal application code, so that makes sense.

> request_async

You mentioned adding a signifier like `async` in the original design meeting as well. My personal opinion is that _everything_ in this module should be async, so it feels weird to make the async-ness explicit. Seems to indicate that there is a non-async version as well. I'm not against adding it, but it feels noisy to me at the moment.

text/1026-net-module.rst
scotttrinh: > There could be some difference between the status of a response and the state of the requesting task, and such a difference may affect how we retry the request.

Yeah, I meant to communicate that the `ResponseState` was indeed the state of the _task_, but I didn't have the concept of `Task` initially.

> I added an Error state for non-retryable failures, like a bad request, too many retries, timed out in receiving response (server may have accepted the request without us knowing). In this case, we can only retry on Failed state (transitioning to InProgress), or certain Complete cases depending on the response status below.

Yeah, this is an interesting idea. It feels a little like the `Task` would be driving a lot of the logic in this case, which I'm not sure I think is a good idea. I'd rather have the thing that knows about retrying own that logic and know how to inspect a `Task<HttpResponse, TaskError>` and retry the `Task`. I think this is the fundamental disagreement that I was having with @1st1 also. To me, these basic methods for making requests should _not_ have any logic and just know protocol-specific things, rather than "business logic" like things (including retrying, concurrency, etc).

> However, perhaps we don't need an additional HttpStatus now (there is already the status field), especially for the try_request MVP.

Yeah, I agree with this. The `task.response.status` should just be the status code returned by the server.

> I just feel wrong mixing this response status with the task state. Like the task state can be shared between http, smtp or friends, but the response status is protocol-specific.

Absolutely, and this "state" part was meant to be the `Task.state`, so yeah, I'll refactor.

scotttrinh: @fantix ⚡ [`3c588ec` (#95)](https://github.com/edgedb/rfcs/pull/95/commits/3c588ecf17359b0ba2de07aa004d65d0f6d11093)

text/1026-net-module.rst
```diff
@@ -0,0 +1,209 @@
...
+its current state. Each protocol will have its own ``net::Task`` concrete type.
+
+.. code-block:: edgeql
+
+  scalar type net::TaskState extending std::enums<Pending, InProgress, Complete, Failed>;
+
+  abstract type net::Task {
+    required state: net::TaskState;
+    required created_at: datetime;
+  }
```
scotttrinh, line 57: I think we also probably want a `code: int16` and/or `reason: str` to store issues with the actual networking.

fantix: Ah I was thinking about something like:

```
  abstract type net::Task {
    required state: net::TaskState;
    required created_at: datetime;
    multi failed_results: net::Result;
  }

  type net::Result {
    required complete_at: datetime;
    required code: net::ErrorCode;
    required message: str;
    # maybe also link to responses (if received one) in subtypes
  };
```

So that the whole retrying process is auditable. Again, this is for the RFC not the MVP.

scotttrinh: Ahh, I see. I personally like the `Task` part to be a little more atomic, otherwise I feel like the state is spread over the `.state` property and all of the `.failed_results` links. I still think I'm looking at this from a totally different perspective of _not_ wanting retries (and other logic) to live in these types at all. 

In my view:
- A `Task` represents exactly one request, one network attempt, and zero or one responses.
- You can look at the `Task` itself to find out if it failed or succeeded
- if you want to retry a request, you'd spawn a new `Task`.

That feels more analogous to a `Promise` from JS (and I'd imagine a `Future` in async python), so perhaps we're just talking about different levels of abstraction where I'm seeing this as "queue some work for an event loop" instead of seeing this as something that has some baked-in domain logic.

At the moment, I think my own preference around behavior of these side-effecting functions in the `net` module is:

- One call = one attempt at delivery
- Every call returns some object that can be inspected about the state of the request, and maybe at some point later, the response

That means that it's up to the caller, so in the case of the `ext::auth` extension for instance, to know when and how to retry. In the future, we'd back up this logic with some kind of durable queue and domain-specific information about what kinds of requests are retryable under what conditions. For now, the auth extension could retry all network failures or 5XX responses in some fixed manner. But, I feel strongly that having any of that logic in this low-level module feels like it will cause us more trouble in the future. Imagine trying to implement a max-retry count in an extension, but because `net` always retries 5XX, you now how to observe both how many times you call `net::http::request` and the internal retry queue rather than knowing that every `net::http::request` call is exactly one outgoing request attempt. And, then you'd only be able to have a max_retries that is some multiple of the internal retry logic.

scotttrinh: Maybe `Task` -> `Future` or `AsyncResult`?

text/1026-net-module.rst
```diff
@@ -0,0 +1,214 @@
...
+
+  scalar type net::TaskState extending std::enums<Pending, InProgress, Complete, Failed>;
+
+  scalar type net::TaskFailure extending std::enums<NetworkError, Timeout>;
+
+  abstract type net::Task {
+    required state: net::TaskState;
+    required created_at: datetime;
+    failure: tuple<failure: net::TaskFailure, message: str>;
+  }
```
scotttrinh, line 60: @1st1 Is this tuple what you had in mind for the failure enum-based error reporting?

text/1026-net-module.rst
fantix: I see your point now. It all makes sense to me! :+1:

scotttrinh: @fantix @1st1 

Any thoughts on the `Task` vs something else naming here? Does `Task` imply more capabilities than just "some work for an event loop to execute"?

aljazerzen (approved):

text/1026-net-module.rst
```diff
@@ -0,0 +1,223 @@
...
+  scalar type net::RequestState extending std::enums<Pending, InProgress, Complete, Failed>;
+
+  scalar type net::RequestFailure extending std::enums<NetworkError, Timeout>;
+
+``net::RequestState`` represents the current asynchronous state of a request.
+The worker process will update the state of the request as it progresses through
+the various states.
+
+``net::RequestFailure`` represents the types of failure that can occur when
+trying to send the request, not a failure response from the server.
```
aljazerzen, line 60: @scotttrinh 

This could be `net::RequestFailureKind` or just `net::FailureKind`, so `ScheduledRequest.failure.failure` would be `ScheduledRequest.failure.kind`.

PR #97, opened 2024-08-21
RFC 1026: Make it more clear this is async

scotttrinh:
We're just scheduling the request, so make this more clear in the name of the function.


fantix (approved):
❤️

PR #104, opened 2025-01-14
RFC 1028: CLI hooks

vpetrovykh:
A proposal to add migration and fixture hooks to gel projects.

### DIFF
--- a/text/1028-cli-hooks.rst
+++ b/text/1028-cli-hooks.rst
@@ -1,0 +1,159 @@
+::
+
+    Status: Draft
+    Type: Feature
+    Created: 2025-01-14
+    Authors: Victor Petrovykh <victor@edgedb.com>
+
+===================
+RFC 1028: CLI hooks
+===================
+
+This RFC discusses addition of hooks to CLI commands. In order to make it
+easier to setup and maintain some of the things dependent on the schema state
+we can introduce hooks for some of the schema altering commands.
+
+
+Motivation
+==========
+
+There are a number of tools that generate code based on the current schema
+(such as query-building tools, or query reflection tools). They need to be
+re-run after every schema change in order to keep the generated code in sync
+with the database. In larger projects the number of things that may need
+updating after a schema change is likely to grow. Migration hooks could make
+it easier to automate these tasks.
+
+There is also a longstanding need to implement some kind of mechanism for
+creating data fixtures. Unlike the migration hooks, fixtures are more likely
+to be run at project initialization or after a branch wipe command.
+
+
+Hooks Configuration
+===================
+
+All hooks are intended to be used within an active project, not for arbitrary
+CLI commands. Thus if a CLI command receives an explicit ``-I <instance>``
+option, the command should not trigger any project hooks even if the instance
+is linked to a project. Only commands that omit the instance and use the
+current project to determine the target instance are going to trigger hooks.
+
+The hooks will be described in the ``gel.toml`` file in the
+``[project-hooks]`` table. We define ``migration.apply.after``,
+``project.init.after``, ``branch.wipe.after``, and ``branch.switch.after``
+hook keys. The values for them are arrays of strings, where each string is
+going to be executed as a shell command. The naming scheme is intended to
+mimic the command structure to clearly indicate which commands will trigger
+the hooks. Current RFC only introduces hooks to be executed after a given
+command, but the naming scheme supports future "before" hooks as well.
+
+The scripts are executed in the order they appear in the array, sequentially.
+So if you have multiple steps that needs to be executed in a certain order,
+you can put the hook commands in the same order you would have executed them
+in the shell.
+
+All hooks will use the project root directory as the execution directory.
+
+
+Configuration Hooks
+-------------------
+
+This category of hooks are intended for one-time setup of important fixtures
+or application configuration. They are perfect for setting up data fixtures or
+exporting connection strings for non-gel tools to use. There are generally two
+scenarios when this kind of hook might be needed:
+
+1) After ``gel project init`` the ``project.init.after`` hook will be
+   executed. This is good for updating any additional configuration that
+   depends on the project database. It is also a good place for data fixture
+   scripts.
+
+2) After ``gel branch wipe`` the ``branch.wipe.after`` hook will be executed.
+   This may be a good place for restoring the data fixtures. However it is
+   probably unnecessary to run any configuration scripts at this time. This is
+   why this hook does not simply run all the same scripts as the
+   ``project.init.after``.
+
+3) After any ``gel branch`` command that changes the current branch
+   (``switch``, ``rebase``, ``merge``) the ``branch.switch.after`` hook will
+   be executed. This is a good place for updating any configuration scripts
+   (e.g. updating Postgres connection sting). This hook can also be used to
+   keep the source code branch in sync with the database branch. This is
+   probably not a good hook for running data fixtures as the data in branches
+   in unaltered between switches.
+
+If ``project init`` runs migrations, the ``project.init.after`` hook is
+triggered *before* the ``migration.apply.after`` hook. The motivation is that
+whatever is necessary as a one-time project setup is likely to be a
+pre-requisite for migration hooks.
+
+Similarly, ``branch.wipe.after`` and ``branch.switch.after`` hooks are
+executed before the migration hook triggers.
+
+
+Migration Hooks
+---------------
+
+This category of hooks is intended for scripts that sync up the codebase with
+the schema state. They are defined by the ``migration.apply.after`` setting.
+It may be re-compiling the query-builder or reflected individual queries. It
+may be generating an ORM compatibility layer. It could also perform any
+project-specific custom operation.
+
+The effects produced by these hook scripts should be idempotent, so that
+running them several times on the same schema (or nearly same schema with
+trivial changes) is safe.
+
+The hooks should run *only once* per ``migration apply`` command, as long as
+some pending change was applied. They are not intended to run after applying
+every migration file if several migrations are applied by one command.
+
+The ``gel watch`` command will also trigger these hooks (without a need of a
+separate TOML config section). This is due to the fact that the ``watch``
+command conceptually monitors the schema file changes and automatically
+creates and applied migrations. So after the schema file changes are applied
+to the database, the ``migration.apply.after`` hook will be executed.
+
+The ``gel branch`` commands may also trigger these scripts since different
+branches can have different schemas. The ``switch``, ``rebase``, ``merge``,
+and ``wipe`` branch commands all potentially change the current schema. This
+means that the hooks associated with applying schema changes must be executed
+for these commands as well.
+
+
+Design Considerations
+=====================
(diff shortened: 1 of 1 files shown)


text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+every migration file if several migrations are applied by one command.
+
+The ``gel watch`` command will also trigger these hooks (without a need of a
+separate toml config section). This is due to the fact that the ``watch``
+command conceptually monitors the schema file changes and automatically
+creates and applied migrations. So after the schema file changes are applied
+to the database, the ``migration.apply.after`` hook will be executed.
+
+The ``gel branch`` commands may also trigger these scripts since different
+branches can have different schemas. The ``switch``, ``reabse``, ``merge``,
```
beerose, line 84:
```suggestion
branches can have different schemas. The ``switch``, ``rebase``, ``merge``,
```

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+CLI commands. Thus if a CLI command receives an explicit ``-I <instance>``
+option, the command should not trigger any project hooks even if the instance
+is linked to a project. Only commands that omit the instance and use the
+current project to determine the target instance are going to trigger hooks.
+
+The hooks will be described in the ``gel.toml`` file in the
+``[project-hooks]`` table. We define ``migration.apply.after``,
+``project.init.after``, and ``branch.wipe.after`` hook keys. The values for
+them are arrays of strings, where each string is going to be executed as a
+shell command. The naming scheme is intended to mimic the command structure to
```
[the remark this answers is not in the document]
vpetrovykh: Good point. I'll explicitly add that the project directory will be the execution directory for all scripts.

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+clearly indicate which commands will trigger the hooks. Current RFC only
+introduces hooks to be executed after a given command, but the naming scheme
+supports future "before" hooks as well.
+
+Since these are mostly intended to be development aids, hooks will only run on
+*local* deployments. Remote/cloud databases are assumed to need different
+management and sync tools that local ones.
```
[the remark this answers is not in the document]
vpetrovykh: Even if we do, we would want different hooks for the cloud, perhaps?

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+This category of hooks are intended for one-time setup of important fixtures
+or application settings. They are perfect for setting up data fixtures or
+exporting connection strings for non-gel tools to use. There are generally two scenarios when this kind of hook might be needed:
+
+1) After ``gel project init`` the ``project.init.after`` hook will be
+   executed. This is good for updating any  additional configuration that
+   depends on the project database. It is also a good place for data fixture
+   scripts.
+
+2) After ``gel branch wipe`` (or ``database wipe``) the
```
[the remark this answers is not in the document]
vpetrovykh: Fair enough, I'll remove it from being mentioned :)

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+   ``branch.wipe.after`` hook will be executed. This may be a good place for
+   restoring the data fixtures. However it is probably unnecessary to run any
+   configuration scripts at this time. This is why this hook does not simply
```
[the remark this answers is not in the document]
vpetrovykh: It's a good point that branch switching may affect things like connection string... Perhaps this needs an additional category.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
scotttrinh: If you have a cloud instance linked to your project (via https://docs.edgedb.com/cli/edgedb_project/edgedb_project_init#edgedb-cloud), I'd expect that these hooks are run just like a local instance?

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,133 @@
...
+===================
+
+All hooks are intended to be used within an active project, not for arbitrary
+CLI commands. Thus if a CLI command receives an explicit ``-I <instance>``
+option, the command should not trigger any project hooks even if the instance
+is linked to a project. Only commands that omit the instance and use the
+current project to determine the target instance are going to trigger hooks.
+
+The hooks will be described in the ``gel.toml`` file in the
+``[project-hooks]`` table. We define ``migration.apply.after``,
```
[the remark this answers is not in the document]
scotttrinh: Yeah, honestly I wrote this before reading the further information that spells this out explicitly, so really no action needed here. We'll make this super obvious in the docs and examples. 👍

elprans (commented):
There should be some discussion on the input to the hooks, i.e. the environment (e.g `GEL_INSTANCE` containing the name of the instance, etc.).

vpetrovykh:
> There should be some discussion on the input to the hooks, i.e. the environment (e.g `GEL_INSTANCE` containing the name of the instance, etc.).

I don't think we need any extra environment variables for the hooks. You could use whatever you would have used if you ran the scripts by hand from the shell. For example, to get the instance name ``gel project info --instance-name``. The idea is that the scripts wouldn't be relying on any hook-specific magic and you could run them (and thus debug them) by hand with the identical effects.

elprans (approved):

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,159 @@
...
+Hooks Configuration
+===================
+
+All hooks are intended to be used within an active project, not for arbitrary
+CLI commands. Thus if a CLI command receives an explicit ``-I <instance>``
+option, the command should not trigger any project hooks even if the instance
+is linked to a project. Only commands that omit the instance and use the
+current project to determine the target instance are going to trigger hooks.
+
+The hooks will be described in the ``gel.toml`` file in the
```
aljazerzen, line 41: Can the table be named just `hooks` instead of `project-hooks`?

Currently, this project manifest contains only two options: `project.schema-dir` and `instance.server-version`.

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,159 @@
...
+``[project-hooks]`` table. We define ``migration.apply.after``,
+``project.init.after``, ``branch.wipe.after``, and ``branch.switch.after``
+hook keys. The values for them are arrays of strings, where each string is
+going to be executed as a shell command. The naming scheme is intended to
```
aljazerzen, line 45: Do we want these commands to be executed as a shell command, or restrict it further to a process invocation?

I.e. do we want to support `command1 && command2 | command3`?

I'd say we really don't need a shell, since that would bring compatibility problems between different OSs and shells.

aljazerzen: We already plan to have *arrays* of commands, which is the same as `command1 && command2 && command3` (sidenote: we probably want to abort our action if any of the hooks returns non-zero status code, right?).

Also, I users really want a shell, they can use it like this:

```
[hooks]
branch.switch.after = ["/bin/bash -c 'echo hello | wc'"]
```

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,159 @@
...
+   scripts.
+
+2) After ``gel branch wipe`` the ``branch.wipe.after`` hook will be executed.
+   This may be a good place for restoring the data fixtures. However it is
+   probably unnecessary to run any configuration scripts at this time. This is
+   why this hook does not simply run all the same scripts as the
+   ``project.init.after``.
+
+3) After any ``gel branch`` command that changes the current branch
+   (``switch``, ``rebase``, ``merge``) the ``branch.switch.after`` hook will
```
aljazerzen, line 78: `branch rebase` and `branch merge` don't change the current branch. They do apply new migrations.

text/1028-cli-hooks.rst
scotttrinh: Yeah, I agree we should be explicit about what environment we actual are running these commands in and how each command does or does not affect the "downstream" commands. I think we can lean on prior art here, though.

In JS-world, you can define a _single_ command in a package.json that will run in a spawned process inherited from your existing shell. Users need to orchestrate running multiple commands themselves explicitly, or if it's complex enough, just invoke some shell script.

In GitHub actions `run` blocks, [they say this](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun):

> Each run keyword represents a new process and shell in the runner environment. When you provide multi-line commands, each line runs in the same shell.

I'm not sure if that's precise enough to answer the question, but I would expect us to match that behavior if we want to stick with "multiple commands"? Maybe we just need to allow a single string, which in TOML can be multi-line, and it's passed exactly to the spawned shell?

text/1028-cli-hooks.rst
aljazerzen: @vpetrovykh 

The question is what's the intended purpose of having switch hook run on `merge` and `rebase`?

text/1028-cli-hooks.rst
vpetrovykh: Given that the toml file is for projects, we can just name the table `hooks`, I guess. Everything here is for the project.

text/1028-cli-hooks.rst
scotttrinh: ahh, so `rebase` and `merge` would not trigger any kind of hook, right? You'd end up triggering a hook when you actually apply the new migrations, but it'd be the `migrations.apply.after` hook?

vpetrovykh: I think you're right in pointing out that those commands don't switch a branch, so they shouldn't run the switch hook. I got a bit overzealous there. They do change the schema migration state, so they should still trigger the migration hook.

aljazerzen: Yes, migrations.apply would trigger.

text/1028-cli-hooks.rst
aljazerzen: I don't think this is very similar to github actions, they have a full syntax built around pipelines of commands.

But it is very, very similar to npm scripts. And npm uses `/bin/sh` and `cmd.exe`: https://docs.npmjs.com/cli/v9/commands/npm-run-script#script-shell

So I guess we should too. And then it does make sense to have a single string instead of array of string, as victor suggests.

text/1028-cli-hooks.rst
```diff
@@ -0,0 +1,159 @@
...
+1) After ``gel project init`` the ``project.init.after`` hook will be
+   executed. This is good for updating any additional configuration that
+   depends on the project database. It is also a good place for data fixture
+   scripts.
+
+2) After ``gel branch wipe`` the ``branch.wipe.after`` hook will be executed.
+   This may be a good place for restoring the data fixtures. However it is
+   probably unnecessary to run any configuration scripts at this time. This is
+   why this hook does not simply run all the same scripts as the
+   ``project.init.after``.
```
aljazerzen, line 75: @vpetrovykh Can you confirm that we also want `branch.wipe.after` to trigger after a `gel database wipe`?

PR #105, opened 2025-01-24
Followup for CLI hooks

aljazerzen:
- **branch merge/rebase does not switch**
- **[project-hooks] -> [hooks]**
- **minor reword**
- **array of strings, /bin/sh, status codes**

PR #106, opened 2025-01-28
Update to CLI hooks 1028 (watch scripts)

vpetrovykh:
Add "watch" configuration in addition to "hooks". This is for monitoring changes to things like files or database config and triggering scripts in response.

### DIFF
--- a/text/1028-cli-hooks.rst
+++ b/text/1028-cli-hooks.rst
@@ -3,7 +3,8 @@
     Status: Draft
     Type: Feature
     Created: 2025-01-14
-    Authors: Victor Petrovykh <victor@edgedb.com>
+    Authors: Victor Petrovykh <victor@edgedb.com>,
+             Aljaž Mur Eržen <aljaz@edgedb.com>
 
 ===================
 RFC 1028: CLI hooks
@@ -164,16 +165,12 @@ of response is possible: trigger when a change to the watched entity is
 detected.
 
 The watch configuration will be described in the ``gel.toml`` project manifest
-file in the ``[watch]`` table. We will introduce ``[watch.<name>]`` sub-tables
-for watching file system changes and triggering scripts.
-
-Each ``[watch.<name>]`` sub-table can have any arbitrary ``<name>`` and
-represents a single watch trigger. Multiple watch triggers can be specified as
-long as they have their own unique ``<name>``.
+file in the ``[[watch]]`` table array. Each ``[[watch]]`` element represents a
+single watch trigger.
 
 Each watch trigger should have the following structure:
 
-* ``glob = "<path-string>"`` - specify file(s) being watched; required
+* ``files = ["<path-string>"]`` - specify file(s) being watched; required
 * ``script = "<command>"`` - command to be executed by the shell; required
 
 We may add more configuration options in the future to the watch triggers to
@@ -192,7 +189,9 @@ This is a backwards incompatible change compared to how ``edgedb watch`` operate
 now.
 
 The output of the scripts will then appear in the same TTY as the ``gel
-watch`` command.
+watch`` command. The output should be prefixed by the ``files`` value so that
+we can indicate which trigger produced the output when several scripts run in
+parallel.
 
 
 Trigger Execution
@@ -211,36 +210,18 @@ watch trigger monitoring several files which have been modified. We want to
 reduce unnecessary repeated execution of the same trigger script by debouncing
 multiple file system change events into a single trigger event.
 
-To achieve this ``gel watch`` should keep an event queue for triggered
-scripts. Any time a new trigger change is detected, the event queue is
-searched for the same trigger event. If it is found, the stale event is
-removed from the queue. The *new* event is added to the end of the queue with
-the current timestamp. This means that the events in the queue appear in
-chronological order. The head of the queue is checked periodically by ``gel
-watch`` and if the timestamp of the trigger event is more than 100ms old, the
-event is *popped* from the queue and the corresponding script is executed.
-This guarantees that no more than one copy of any trigger event is ever
-present in the queue at the same time. It also guarantees that eventually all
-triggered events will execute as long as updates triggering them actually
-stop.
+By default we want all trigger events to have a 100ms debounce delay before
+executing. This value may be configurable in the future and be different for
+each trigger.
 
 Another consideration affecting debouncing is whether the same trigger script
 is already currently running. It means that ``gel watch`` must keep track of
 all the currently running scripts so that when a new script from the trigger
 queue becomes eligible to run an additional check can be made. If there is
 already a copy of the same trigger event script running, the newly eligible
-copy is auto-debounced (by 100ms) and moved to the end of the trigger queue.
-This ensured that only one copy of any trigger script is actually running at
-the same time, further reducing unnecessary churn.
-
-In general, it makes sense to execute triggered scripts in parallel. This
-reduces the latency in case some scripts are slow-running. However, sometimes
-(e.g. when scripts overlap in which files they change) it is desirable that no
-two scripts are running at the same time within the project. We can add
-``--serial`` flag to ``gel watch`` command in order force trigger scripts to
-be executed one at a time. Note that if there are long-running scripts the may
-be a significant delay between the triggering event and the actual script
-execution if ``--serial`` mode is enabled.
+copy is auto-debounced while the previous copy is still running. This ensures
+that only one copy of any trigger script is actually running at the same time,
+further reducing unnecessary churn.
 
 Another consequence of executing triggered scripts in the background is that
 sometimes the changes are invalid in some way and the script will fail. This
@@ -252,8 +233,8 @@ non-recoverable state.
 Files
 -----
 
-The ``[watch.<name>]`` sub-table should contain at least ``glob`` and
-``script`` keys. The ``glob`` value is interpreted as a file system path. It
+The ``[[watch]]`` elements should contain at least ``files`` and
+``script`` keys. The ``files`` value is an array of file system paths. It
 can also contain common glob patterns (such as provided by `this library
 <https://docs.rs/globset/latest/globset/#syntax>`_).
 
@@ -272,8 +253,8 @@ the scripts are always executed in WSL.
 
 An example of this configuration::
 
-    [watch.files]
-    glob = "queries/*.edgeql"
+    [[watch]]
+    files = ["queries/*.edgeql"]
     script = "npx @edgedb/generate queries"
 
 Only files in the project directories can be watched. If the ``gel.toml``
@@ -319,26 +300,26 @@ prevent accidental triggering of the hooks in production.
 Future Possibilities
 ====================
 
-The debouncer delay can be configurable in the ``[watch]`` section as
+The debouncer delay can be configurable in the ``[[watch]]`` section as
 ``deboucer-delay=<integer>`` with the value being the delay in milliseconds.
 This means that it's no longer sufficient to check the head of the trigger
 queue to find the script to be executed. The entire queue must be checked and
 all triggers that are older than their delay time must be executed.
 
-We can introduce a ``[watch.gel-config]`` sub-table for monitoring changes to
+We can introduce a ``gel-config`` to ``[[watch]]`` spec for monitoring changes to
 the various database config values and responding to them. This would make
-``glob`` an optional key and add ``gel-config = <string>`` as a way to specify
+``files`` an optional key and use ``gel-config = <string>`` as a way to specify
 which config variable we are watching.
 
 We can set up a logfile for watch scripts (because that creates a record which
 survives closing of terminals or reboots) as an additional convenience
-feature. This can be specified in the general ``[watch]`` section as
+feature. This can be specified in the general ``[watch-global]`` section as
 ``logfile="<path-to-logfile>"``.
 
 The process launched by a script could have additional environmental variables
 such as ``GEL_HOOK_NAME`` with value, for example, ``branch.wipe.before``.
 
-We can add the following optional keys to the ``[watch.<name>]`` trigger sub-tables:
+We can add the following optional keys to the ``[[watch]]`` element:
 
 * ``retry-delay = <integer>`` - retry delay in milliseconds
 * ``retry-attempts = <integer>`` - number of retry attempts
@@ -354,10 +335,19 @@ both of the keys are missing, no retry attempt should be made and the keys'
 
 An example of a valid retry config::
 
-    [watch.remote_data]
-    glob = ".env"
+    [[watch]]
+    files = [".env"]
     script = "./sync_with_remote.sh"
     retry-delay = "5s"
+
+In general, it makes sense to execute triggered scripts in parallel. This
+reduces the latency in case some scripts are slow-running. However, sometimes
+(e.g. when scripts overlap in which files they change) it is desirable that no
+two scripts are running at the same time within the project. We can add
+``--serial`` flag to ``gel watch`` command in order force trigger scripts to
+be executed one at a time. Note that if there are long-running scripts the may
+be a significant delay between the triggering event and the actual script
+execution if ``--serial`` mode is enabled.
 
 
 Rejected Ideas
@@ -415,12 +405,12 @@ it's expected to be committed as part of a repository.
 We no longer want to use a minimalistic "glob=script" setup for files being
 watched. This format is hard to extend in the future if any additional config
 options are deemed necessary. Instead we will use sub-tables
-``[watch.<name>]`` for each watch trigger, making it easier to add custom
+``[[watch]]`` for each watch trigger, making it easier to add custom
 options on a per-trigger basis. The simplest watch example is still quite
 short and arguably more easier to understand::
 
-    [watch.files]
-    glob = "queries/*.edgeql"
+    [[watch]]
+    files = ["queries/*.edgeql"]
     script = "npx @edgedb/generate queries"
 
 Watch should not have a ``--retry-sec`` flag for controlling retry behavior
--- a/text/1028-cli-hooks.rst
+++ b/text/1028-cli-hooks.rst
@@ -28,6 +28,9 @@ There is also a longstanding need to implement some kind of mechanism for
 creating data fixtures. Unlike the migration hooks, fixtures are more likely
 to be run at project initialization or after a branch wipe command.
 
+Also, there is a need to watch for file changes (schema, queries) and
+potentially respond to these by running some scripts.
+
 
 Hooks Configuration
 ===================
@@ -39,15 +42,25 @@ is linked to a project. Only commands that omit the instance and use the
 current project to determine the target instance are going to trigger hooks.
 
 The hooks will be described in the ``gel.toml`` project manifest file in the
-``[hooks]`` table. We define ``migration.apply.after``,
-``project.init.after``, ``branch.wipe.after``, and ``branch.switch.after``
-hook keys. Their values are strings that are going to be executed as shell
-commands.
+``[hooks]`` table. We define the following hook keys:
+
+* ``project.init.before``
+* ``project.init.after``
+* ``branch.switch.before``
+* ``branch.switch.after``
+* ``branch.wipe.before``
+* ``branch.wipe.after``
+* ``migration.apply.before``
+* ``migration.apply.after``
+* ``schema.update.before``
+* ``schema.update.after``
+
+Their values are strings that are going to be executed as shell commands.
 
 The naming scheme is intended to mimic the command structure to clearly
-indicate which commands will trigger the hooks. Current RFC only introduces
-hooks to be executed after a given command, but the naming scheme supports
-future "before" hooks as well.
+indicate which commands will trigger the hooks. The exception to that is the
+``schema.update`` hooks that are intended to trigger in response to any kind
+of schema change, regardless of which command causes it.
 
 All hooks will use the project root directory as the execution directory.
 Hooks are executed using ``/bin/sh`` on all platforms. On Windows, the hooks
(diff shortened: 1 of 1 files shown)


text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,105 @@ means that the hooks associated with applying schema changes must be executed
...
+to the various database config values and responding to them.
+
+The general structure of watch tables is going to use the key (potentially
+quoted) to specify *what is being watched* and the corresponding value to
+specify the script that will be triggered by changes.
+
+Since the triggered scripts will be executed in response to changes, we cannot
+meaningfully display the output of the scripts to the user. We may want to
+setup an optional logfile to record the script output. This can be specified
+in the general ``[watch]`` section as ``logfile="<path-to-logfile>"``.
```
scotttrinh, line 155: I'm not sure I understand this: why wouldn't you see the output in the same TTY as the running watch command?

text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,105 @@ means that the hooks associated with applying schema changes must be executed
...
+
+We may want to setup a bit of a delay before triggering the scripts so as to
+allow a sequence of changes to happen before the response script is triggered.
+This is mostly to reduce unnecessary multiple triggers for the same watched
+entity.
```
scotttrinh, line 160: Yeah, I was wondering a bit about overlapping globs like `**/*.edgeql` and `dbschema/*.edgeql` for instance: We don't necessarily want to try to sequence the scripts in a way that they affect each other, so blind parallel execution seems reasonable. Users will need to take care that their scripts do not cause conflicts with each other.

text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,105 @@ means that the hooks associated with applying schema changes must be executed
...
+applying schema changes without writing a migration step for each such change.
+Perhaps a ``gel migration try`` command that would attempt to apply changes
+based on the current schema files without writing a corresponding migration
+file.
+
+With that CLI change we could make ``gel watch`` command create the following
+entry in ``gel.toml``::
+
+    [watch.files]
+    "dbschema/**.schema"="gel migration try"
```
1st1, line 213:
```suggestion
    "dbschema/**.gel"="gel migration try"
```

I kinda like the generalization idea but I think it's a step too far. IMO `gel watch` should just require either or both `--schema` / `--files` flags and do the right thing.

text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,105 @@ means that the hooks associated with applying schema changes must be executed
...
-    branch.switch.after=[
-      "setup_dsn.sh"
-    ]
-    migration.apply.after=[
-      "gel-orm sqlalchemy --mod compat --out compat"
-    ]
+    project.init.after="setup_dsn.sh"
+    branch.wipe.after=""
+    branch.switch.after="setup_dsn.sh"
+    migration.apply.after="gel-orm sqlalchemy --mod compat --out compat"
```
1st1, line 226: Please add an explanation why we decided to go for a single command vs the previous design where we wanted to have a list of them.

aljazerzen: This is a leftover from https://github.com/edgedb/rfcs/pull/105, which was a followup for this: https://github.com/edgedb/rfcs/pull/104#discussion_r1925688396

text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,105 @@ means that the hooks associated with applying schema changes must be executed
...
+Files
+-----
+
+The ``[watch.files]`` table should contain keys that are interpreted as file system paths. The following special symbols should be allowed:
+
+- ``?`` to denote a single character wild card
+
+- ``*`` to denote an arbitrary string of characters in a name
+
+- ``**`` to denote an arbitrary sub-path
```
aljazerzen, line 178: I suggest expanding this to common glob syntax.

Here is a library we can use: https://docs.rs/globset/latest/globset/#syntax

text/1028-cli-hooks.rst
aljazerzen: From reading the code, I see that, currently, `gel watch` is triggering `crate::migrations::dev_mode::migrate`, without having any state in-between invocations, apart of error reporting.

https://github.com/edgedb/edgedb-cli/blob/master/src/watch/main.rs#L154-L182

So I actually do think generalization is possible.

text/1028-cli-hooks.rst
aljazerzen: So there are two problems:

- IDEs or other tools modifying/moving multiple files in one after another. The RFC says (and I agree) that all the events should trigger the watch script only once. To make this happen, we'll use this: https://docs.rs/notify-debouncer-mini/latest/notify_debouncer_mini/

- scripts modifying more files, which could recurse into the watch script again. This would be partially solved by a smart debouncer, but not entirely, since the watch script could first do some slow queries (say a second) and only then write the files. So I think it would be better to allow recursion, as you suggest @scotttrinh

text/1028-cli-hooks.rst
scotttrinh: Fair, but that's commonly solved by actually piping stdout/err, right? I'm fine with a more user-friendly way to configure it like the suggestion here, but I would _hope_ that we'd also see stdout/stderr in the same terminal as watch: I'd be 😱 if I had to tail a log file to see the live output of the scripts.

aljazerzen: Yes, showing output in TTY is a must. Logfile would just be nice.

text/1028-cli-hooks.rst
aljazerzen: @vpetrovykh are we fine with also having `{a,b}`, `[ab]` and `[!ab]` in the glob syntax?

aljazerzen: Ignore this, I see that you've updated with link to the lib.

text/1028-cli-hooks.rst
```diff
@@ -157,4 +244,12 @@ could use whatever you would have used if you ran the scripts by hand from the
...
\ No newline at end of file
+the identical effects.
+
+We don't want to provide a list of scripts to run for hooks and watch
+triggers. This is because with a list of scripts we need to specify what
+happens when some scripts fail. Should the rest of the list be executed or
+aborted. Under different circumstances different approaches would make sense
+and we would need to implement all these interaction variants. Instead any
+such complexity can be handled inside a singe script that the trigger
+references.
```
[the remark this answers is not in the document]
aljazerzen: My interpretation is that all matching scripts are ran (sorted alphabetically, by the glob pattern).

text/1028-cli-hooks.rst
```diff
@@ -122,23 +125,107 @@ means that the hooks associated with applying schema changes must be executed
...
+Files
+-----
+
+The ``[watch.files]`` table should contain keys that are interpreted as file
+system paths. They can also contain common glob patterns (such as provided by
+`this library <https://docs.rs/globset/latest/globset/#syntax>`_).
+
+The paths should be valid in the underlying file system (therefore using ``/``
+for Linux and ``\`` for Windows, etc.). The relative file paths are assumed to
+start at the project root.
```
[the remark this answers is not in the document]
aljazerzen: I'd say no. I don't have a good reason why not though. Apart from being an anti-pattern of "how to organize code into repos".

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
dnwpark: I guess the possible concern is that earlier it says:
`The order in which hook *keys* appear does not impact their priority (we don't want people getting subtle bugs due to different key order).`

Which means we could still be introducing subtle ordering bugs anyways.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+* ``gel project init`` command triggers the ``project.init.before``
+  and ``project.init.after`` hook. If the migrations are applied at the end of
+  the initialization, then the ``migration.apply.before`` and
+  ``migration.apply.before`` hooks are also triggered.
+* ``gel branch switch`` command triggers ``branch.switch.before``,
+  ``schema.update.before``, ``branch.switch.after``, and ``schema.update.after``
+  hooks in that relative order.
+* ``gel branch wipe`` command triggers the ``branch.wipe.before``,
+  ``schema.update.before``, ``branch.wipe.after``, and ``schema.update.after``
+  hooks in that relative order.
```
1st1, line 99: `edgedb branch wipe` doesn't wipe the schema, only date, does it?

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
-project-specific custom operation.
+or application configuration. They are perfect for setting exporting
+connection strings for non-gel tools to use. They are primarily triggered by
+the corresponding CLI commands, but sometimes commands may trigger more than
+one hook:
+
+* ``gel project init`` command triggers the ``project.init.before``
+  and ``project.init.after`` hook. If the migrations are applied at the end of
+  the initialization, then the ``migration.apply.before`` and
+  ``migration.apply.before`` hooks are also triggered.
```
1st1, line 93: You forgot schema.update

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+#. ``branch.wipe.after``
+#. ``migration.apply.after``
+#. ``schema.update.after``
+
+
+Schema Hooks
+------------
+
+These hooks are a bit more abstract than the ones corresponding to specific
+CLI commands. The ``schema.update.before`` and ``schema.update.after`` hooks
```
1st1, line 130: Given that this is a "meta" hook, it seems to me that it's unnecessary for it to have `.before`, it's basically a notification.

1st1: I'd just rename it to `schema.update` (or `schema.updated'? `schema.changed?`) and call it a day. Most likely this will be the most popular hook, so maybe let's make it high-level and shortly named.

text/1028-cli-hooks.rst
1st1: I'd like us to list use cases here.

`project.init.before` can check that you have the necessary tools installed in your system, like `npx`

`project.init.after` can print a welcome message with instructions and next steps.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+the corresponding CLI commands, but sometimes commands may trigger more than
+one hook:
+
+* ``gel project init`` command triggers the ``project.init.before``
+  and ``project.init.after`` hook. If the migrations are applied at the end of
+  the initialization, then the ``migration.apply.before`` and
+  ``migration.apply.before`` hooks are also triggered.
+* ``gel branch switch`` command triggers ``branch.switch.before``,
+  ``schema.update.before``, ``branch.switch.after``, and ``schema.update.after``
+  hooks in that relative order.
```
1st1, line 96: `branch.switch.before` might check if `.gel` files have any uncommitted changes and abort the operation.
`branch.switch.after` might check your `git` branches and remind switching git branch too.

text/1028-cli-hooks.rst
1st1: `branch.wipe.before` might abort the operation.

BTW, @aljazerzen, can we set some environment variables for the hook commands we run? Like for `branch.wipe.before` we can set 

```
env GEL_HOOK_NAME=branch.wipe.before \
    GEL_HOOK_INSTANCE=1st1/production
    GEL_HOOK_BRANCH=main
```

and then people will be able to write scripts that protect/react to things.

We don't need to implement this now, but I'd like the RFC to have a "future extensions" section that mentions this. And we should discuss what happens if a "before" hook errors out (IMO the command needs to be aborted), and what happens if an "after" hook errors out (IMO there's not much we can do, so just face plant).

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+Watch Configuration
+===================
+
+Sometimes in a Gel project there's a need to respond to some file changes: run
+a migration due to schema change or run code generators due to query file
+change. The CLI already supports watching the schema for changes and applying
+them, but we can generalize this mechanism and make it more flexible, much
+like the hooks that respond to CLI commands.
+
+The idea is to have a mechanism for specifying what changes you want to watch
```
1st1, line 159:
```suggestion
The idea is to have a mechanism for specifying a file system path you want to watch for modifications
```

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+and a script that gets triggered by those changes. Unlike hooks, there are no
+"before" and "after" triggers here. Only one kind of response is possible:
+trigger when a change to the watched entity is detected.
+
+The watch configuration will be described in the ``gel.toml`` project manifest
+file in the ``[watch]`` table. We will introduce a ``[watch.files]`` sub-table
+for watching file system changes and triggering scripts. Potentially, we will
+also later introduce a ``[watch.gel-config]`` sub-table for monitoring changes
```
1st1, line 167: Move ``[watch.gel-config]`` to "future extensions" or something. Probably should be just ``[watch.config]``? Or ``[watch.server-config]``?

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+specify the script that will be triggered by changes.
+
+The ``gel watch`` command (without needing further options) would then be used
+to start the watch process and monitor whatever is specified in the
+``gel.toml``. By default ``gel watch`` will only watch the files specified in
+the ``gel.toml`` config and execute the trigger scripts. Running ``gel watch
+--migrate`` will additionally monitor schema changes and perform real-time
+migrations.
+
+This is a backwards incompatible change compared to how ``gel watch`` operates
```
1st1, line 181:
```suggestion
This is a backwards incompatible change compared to how ``edgedb watch`` operates
```

text/1028-cli-hooks.rst
scotttrinh: > `edgedb branch wipe` doesn't wipe the schema, only data, does it?

It actually does `reset schema to initial;` which does wipe the schema itself along with the data. FWIW, it would be awesome if there was a data-only version of this command since manually "clearing" a branch is actually a bit annoying.

text/1028-cli-hooks.rst
scotttrinh: >  it seems to me that it's unnecessary for it to have .before, it's basically a notification.

I think `schema.update` has just as much reason to have "before" hooks, especially if it can abort the update.

> Most likely this will be the most popular hook, so maybe let's make it high-level and shortly named.

I think it would be very strange if one set of hooks was named in a totally different way than the rest to save a few keystrokes. If we want to just have a small improvement we can save a few characters by going with `pre` and `post` instead? Or do a similar imperative<>past-tense scheme for the other hooks (`migration.apply` and `migration.applied` etc)?

But my feeling is that most people will end up with:

```toml
[hooks]
"schema.update.after" = "npx generate edgeql-js"
```

And that doesn't seem much worse than

```toml
[hooks]
"schema.updated" = "npx generate edgeql-js"
```

And _especially_ if they do the `project.init.after` hook, this feels really weird:

```toml
[hooks]
"schema.updated" = "npx generate edgeql-js"
"project.init.after" = "npx tsx ./initialize-db.ts"
```

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+now.
+
+The output of the scripts will then appear in the same TTY as the ``gel
+watch`` command. We may also consider setting up a logfile (because that
+creates a record which survives closing of terminals or reboots) as an
+additional convenience feature. This can be specified in the general
+``[watch]`` section as ``logfile="<path-to-logfile>"``.
```
scotttrinh, line 188: This should move to a "future extensions" section as well, I think.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+
+We may want to setup a debouncer so that we can delay before triggering the
+scripts on a sequence of changes. This is mostly to reduce unnecessary
+multiple triggers for the same watched entity. The debouncer delay may be
+configured in the ``[watch]`` section as ``deboucer-delay=<integer>`` with the
+value being the delay in milliseconds.
```
scotttrinh, line 194: Debouncing seems like a must-have feature, but being able to tweak the timing seems like a nice-to-have one.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
scotttrinh: I think one counter-argument to this is that sometimes the location of the `gel.toml` file will not be the root of the logical "project". Sometimes people really want to set up a monorepo with one of the sub-projects "owning" the database, so having the `gel.toml` in the true monorepo root feels _less_ encapsulated. And in these cases, there might be valid use cases for reaching to a sibling package/module to trigger some script.

One possible example is having a monorepo with a Python backend (that owns the database) and having a separate single-page app and wanting to run some tests on the backend when changes to the frontend happen.

---

FWIW, I think in this particular scenario, you're now using the `gel watch` feature as a very generic file-watcher utility like watchman, so this might not be a use case we care about. We can stick to a stricter behavior now and gather feedback and use cases.

text/1028-cli-hooks.rst
```diff
@@ -157,4 +278,30 @@ could use whatever you would have used if you ran the scripts by hand from the
...
+happens when some scripts fail. Should the rest of the list be executed or
+aborted. Under different circumstances different approaches would make sense
+and we would need to implement all these interaction variants. Instead any
+such complexity can be handled inside a singe script that the trigger
+references.
+
+We no longer try to generalize the monitoring the schema file changes and
+auto-migrating them as part of regular ``gel.toml`` watch spec. There is a bit
+of special handling and safe-guards involved in that monitoring that make it a
+little too special. Instead we offer ``--monitor`` flag to enable
```
scotttrinh, line 294:
```suggestion
little too special. Instead we offer ``--migrate`` flag to enable
```
right?

text/1028-cli-hooks.rst
vpetrovykh: It does wipe the schema, not just data.

text/1028-cli-hooks.rst
vpetrovykh: The use cases for some of the `before` hooks are more contrived than others, but excluding some hooks from having a `before` option seems to be too harsh. Checking stuff before proceeding with the command is a pretty general argument for any hook. "I want to check something before *any* schema change" and "I specifically want to check stuff before a migration" are both valid.

text/1028-cli-hooks.rst
```diff
@@ -39,15 +42,25 @@ is linked to a project. Only commands that omit the instance and use the
...
+
+Their values are strings that are going to be executed as shell commands.
 
 The naming scheme is intended to mimic the command structure to clearly
-indicate which commands will trigger the hooks. Current RFC only introduces
-hooks to be executed after a given command, but the naming scheme supports
-future "before" hooks as well.
+indicate which commands will trigger the hooks. The exception to that is the
+``schema.update`` hooks that are intended to trigger in response to any kind
+of schema change, regardless of which command causes it.
```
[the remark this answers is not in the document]
vpetrovykh: I do talk about the ``schema.update`` hooks in their own section, where I mention this stuff.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,184 @@ are always executed in WSL.
...
+* ``gel branch rebase`` and ``gel branch merge`` commands trigger
+  ``migration.apply.before``, ``schema.update.before``,
+  ``migration.apply.after``, and ``schema.update.after`` hooks in that
+  relative order. Notice that although these are branch commands, but they do
+  not change the current branch, instead they modify and apply migrations.
+  That's why they trigger the ``migration.apply`` hooks.
+* ``gel migrateion apply`` (or ``gel migrate``) command triggers
+  ``migration.apply.before``, ``schema.update.before``,
+  ``migration.apply.after``, and ``schema.update.after`` hooks in that
+  relative order.
```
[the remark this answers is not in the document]
vpetrovykh: I've added a note about this into the `Rejected Ideas` section. The TLDR version is that "branch switch causes schema change" and "migration causes schema change" may need different things (like Postgres connection string) setup before the schema change. Hence why order is what it is.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
aljazerzen: I don't want it:
- being random (which would be caused by storing them in a HashMap),
- depend on the order in the TOML table (because TOML tables are not supposed to have an order).

So I picked "an order" which might change with releases of the CLI, but will not change with each invocation of `gel watch` or with each file-change-event.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
aljazerzen: Yeah, this does not need to be mentioned here, because there are many ways in which migration might be applied:

- gel migration apply
- gel watch
- gel project init
- gel branch rebase
- gel branch merge

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,182 @@ are always executed in WSL.
...
+
+The output of the scripts will then appear in the same TTY as the ``gel
+watch`` command.
+
+Multiple consecutive file system change events will be debounced into a single
+event. For example, given scripts ``src/**`` and ``*.py`` and writes to
+``src/a.py``, ``src/b.py`` and ``src/c.rs``, both scripts will trigger exactly
+once. To achive this, ``gel watch``` will wait for a small amount of time
+(100ms) after receiving first event before executing matching scripts. This is
+mostly to reduce unnecessary multiple triggers for the same watched entity.
```
aljazerzen, line 192: @vpetrovykh I've updated this a bit, does it make sense?

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,182 @@ are always executed in WSL.
...
+is considered part of normal operation (e.g. syntax error in a saved query
+file) and the scripts should be such that failure does not create some
+non-recoverable state.
+
+Due to the nature of monitoring changes, we cannot guarantee any particular
+order in which the watch scripts will be triggered when multiple changes occur
+at once. This is independent of the source of multiple triggers: whether it is
+due to multiple files being updated or multiple watch rules matching the same
+file. The scripts are effectively triggered asynchronously and the exact order
+is an implementation detail that should not be relied upon.
```
aljazerzen, line 205: This differs from current implementation:
```python
changed_paths = wait_for_changes()

for matcher in matchers:
  is_matched = any(matcher.matches(p) for p changed_paths)
  if is_matched:
    await run_script(matcher.script)
```

Differences:
- the order of fs events does not matter,
- order of matchers (and executed scripts) is fixed between runs, but it is an implementation detail,
- scripts are run sequentially (I wait for first to complete before running the next one).

Is this behavior ok? It can be changed if we have a good reason to do so.

aljazerzen: @vpetrovykh

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,182 @@ are always executed in WSL.
...
+
+Files
+-----
+
+The ``[watch.files]`` table should contain keys that are interpreted as file
+system paths. They can also contain common glob patterns (such as provided by
+`this library <https://docs.rs/globset/latest/globset/#syntax>`_).
+
+The paths should be valid in the underlying file system (therefore using ``/``
+for Linux and ``\`` for Windows, etc.). The relative file paths are assumed to
```
aljazerzen, line 216: This is really inconvenient because it makes `gel.toml` incompatible between OSs. Can we make it always use `/`, even on Windows?

text/1028-cli-hooks.rst
```diff
@@ -149,6 +268,26 @@ mechanism for distinguishing cloud *development* and *production* instances to
...
+We can set up a logfile for watch scripts (because that creates a record which
+survives closing of terminals or reboots) as an additional convenience
+feature. This can be specified in the general ``[watch]`` section as
+``logfile="<path-to-logfile>"``.
+
+The process launched by a script could have additional environmental variables
+such as `GEL_HOOK_NAME` with value, for example, `branch.wipe.before`.
+
+Watch could have a `--retry-sec` flag, which would retry running a script
+if it had exited with a non-zero status code.
```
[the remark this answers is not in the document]
aljazerzen: I've implemented `--retry-sec` because that used to be the behavior of `edgedb watch` (delay was set to 10sec).

With current implementation, it is most useful when there are no `[watch.files]` in the manifest.

I do agree that in general, it should be a per-script setting.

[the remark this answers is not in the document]
vpetrovykh: I can update the RFC to introduce a richer way to describe individual watch settings. It sounds like it might be a good idea given that we tend to add features and with the minimalistic setup adding a new feature is hard.

However, what was a global `--retry-sec` option doing for a `gel watch` without any `[watch.files]`? By default the only thing that could be watched is the schema (currently activated with `--migrate`). But applying migrations is a *deterministic* process. Is that retry intended to specifically help with applying migrations to a remote instance which could have connection errors? If yes, then under what conditions would we not want to always retry on network error? It doesn't seem like for watching the schema it makes sense to use the "retry on error" mechanism and instead we should keep doing our magic and retry only on network errors automatically.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,182 @@ are always executed in WSL.
...
+file. The scripts are effectively triggered asynchronously and the exact order
+is an implementation detail that should not be relied upon.
 
-The ``gel branch`` commands may also trigger these scripts since different
-branches can have different schemas. The ``switch``, ``rebase``, ``merge``,
-and ``wipe`` branch commands all potentially change the current schema. This
-means that the hooks associated with applying schema changes must be executed
-for these commands as well.
+
+Files
```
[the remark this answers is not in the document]
vpetrovykh: > Another take on describing the execution model:
> 
> * on changes, all matching scripts are triggered in parallel,
> * during execution, watch is not suspended and new changes trigger immediately trigger matching scripts,
> * if a script is already executing, a new invocation of the same script waits for the current one to finish.
> 
> I'd add that "at most one invocation" can be waiting for current script to finish. All other are discarded. For example, a script is started and then we receive 3 separate events for the same scripts. When the first script invocation finishes, it is re-run, but only once, even though we received 3 events for it.

I think this makes sense as the default behaviour and if we later discover that users want the option to cancel stale scripts, we can add that as part of the toml config options rather than a blanket CLI flag.

text/1028-cli-hooks.rst
vpetrovykh: Yeah, I can phrase it to be always `/`. The compatibility issue is important.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
aljazerzen: >  what was a global --retry-sec option doing

There wasn't a flag, it was just an unconditional behavior of how watch worked. And it makes sense: if you updated the schema, but the network was down, the `migration apply --dev` would retry every 10sec until success.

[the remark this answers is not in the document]
aljazerzen: Re updating the RFC: the question is "would it be useful to have more settings per-script" / "do we want to future-proof the manifest".

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
aljazerzen: Alright, makes sense. I'll change the default to not suspend during execution. Should I change the flag to be named `--suspend-during-execution`? `--no-recurse`?

---

And also, are we sure we need parallel execution?

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
vpetrovykh: Adding a per-script retry option to watch makes sense because some of the watch scripts may be doing things over the network. It makes sense to give users a way to configure retries for such occasions. This is a lot easier to do in the manifest than via the CLI options. This feature alone is a strong indication that we want to have room for customizing how watch triggers work, so future-proofing seems very reasonable to me.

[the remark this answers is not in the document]
vpetrovykh: > > what was a global --retry-sec option doing
> 
> There wasn't a flag, it was just an unconditional behavior of how watch worked. And it makes sense: if you updated the schema, but the network was down, the `migration apply --dev` would retry every 10sec until success.

OK, makes sense, so we should keep that as implicit behavior of `--migrate` and remove the `--retry-sec` altogether. If the retry option is needed it is needed on a script-by-script basis.

text/1028-cli-hooks.rst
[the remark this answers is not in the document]
vpetrovykh: > Alright, makes sense. I'll change the default to not suspend during execution. Should I change the flag to be named `--suspend-during-execution`? `--no-recurse`?

I think that suspending watching is _never_ a good idea. The meaningful distinction is parallel vs serialized execution.

If you suspend watching, then the implementation detail of the order of triggered scripts becomes extremely important. E.g. if I "save all changes" what happens? 
- Do all of them get detected on the same watch tick and multiple watch targets will be triggered?
- Do some of the changes get detected first and then _block_ other changes from being watched?
- Would I get slightly different behavior from using my key shorcuts to switch tabs and save 2-3 files in quick succession?

If you never suspend watching, then the results for the above are all the same - all changes will be eventually noticed and all triggers will fire. If you suspend watching, then the actual trigger pattern will be sensitive to how exactly files were saved.

I think it is a mistake to fear recursion and try to mitigate it because mitigation introduces subtle differences in behavior for everyone, while recursion would only maybe affect a small number of users who try to do that.

text/1028-cli-hooks.rst
```diff
@@ -56,70 +69,230 @@ are always executed in WSL.
...
+searched for the same trigger event. If it is found, the stale event is
+removed from the queue. The *new* event is added to the end of the queue with
+the current timestamp. This means that the events in the queue appear in
+chronological order. The head of the queue is checked periodically by ``gel
+watch`` and if the timestamp of the trigger event is more than 100ms old, the
+event is *popped* from the queue and the corresponding script is executed.
+This guarantees that no more than one copy of any trigger event is ever
+present in the queue at the same time. It also guarantees that eventually all
+triggered events will execute as long as updates triggering them actually
+stop.
```
aljazerzen, line 225: This might be too specific: it is describing implementation, not the behavior. And currently, it is not implemented with a single event queue.

vpetrovykh: OK, I'll rewrite it as more abstract behavior. Just need to emphasise that the goal is to never have more than one copy of a script waiting in "debouncing" mode.

PR #107, opened 2025-01-30
Update branches RFC #1025

vpetrovykh:
Set up instance-specific and project-specific way to specify the current branch. Both of them de facto exist, but they were not coordinated resulting in bugs.

### DIFF
--- a/text/1025-branches.rst
+++ b/text/1025-branches.rst
@@ -85,8 +85,8 @@ database states that correspond to the schema and avoiding accidental
 mismatches in migrations.
 
 
-Project and CLI Integration
----------------------------
+Branch Configuration
+--------------------
 
 Currently the rest of our tools rely on ``$CONFIG/credentails`` to find
 default credential settings for projects. That includes the "database" (which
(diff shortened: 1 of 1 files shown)


vpetrovykh:
We have 2 ways of specifying branches:

1) As part of the instance credentials
2) As part of project configuration

Our branch CLI tools all work with the instance credentials to keep track of the current branch. However, linking a cloud instance to a project allows to override the branch associated with the cloud instance via a different mechanism. In addition to the instance config we have a *project* config that allows to override `instance-name`, `cloud_profile`, `project-path`, and `database`. This mechanism predates our branches concept.

The `database` setting was introduced by https://github.com/edgedb/edgedb-cli/pull/1086 in order to allow multiple projects to be linked to the same cloud instance, but using different databases.

Unfortunately this way of setting the current branch is buggy in our CLI tools. Which leads to a situation where CLI branch tools work fine for local instances, but adding and the removing a cloud instance may cause `database` to be overridden. The CLI tools then read this `database` setting as the current branch, but they don't *update* it when switching branches, so the branch is stuck.

vpetrovykh:
There is no reason why projects with local and cloud instances should treat branches differently. So we need to make them use the settings consistently.

It seem that we have 2 scenarios:

1) branch commands targetting an instance

   If a branch command has an explcit `-I` option, it should refer to and modify the instance credentials (ignoring project `database` setting).

2) project branch commands

   If a branch command omits `-I` and relies on the project settings, it must never modify the instance credentials, instead updating the `branch` (or `database` when already present) in the project settings (regardless of whether this is a local or cloud instance).

PR #109, opened 2025-03-24
1029: Role Based Access Control

aljazerzen:


text/1029-rbac.rst
```diff
@@ -0,0 +1,152 @@
...
+        CREATE ACCESS POLICY allow_warehouse
+            ALLOW SELECT
+            USING (sys::get_current_role() = 'warehouse_exporter');
+
+        CREATE ACCESS POLICY allow_inserts
+            ALLOW INSERT;
+    };
+
+
+.. access policies that return an error when an object is accessed?
```
msullivan, line 122: Maybe using `assert` could accomplish that?

text/1029-rbac.rst
```diff
@@ -0,0 +1,152 @@
...
+
+.. access policies for functions such as sys::reset_query_stats?
```
msullivan, line 124: Yeah, and we want to be able to restrict *reading* the stats also.

msullivan (commented):
This is mostly good, but I think we need to be able to cover more cases. In particular, we need to support:
 - Restricting access to functions like `sys::reset_query_stats` and (probably) `net::http::schedule_request`
 - Restricting access to certain standard *objects*, like the query stats

We could have built-in capabilities for each of the things we care about here, but I think it probably does make sense to bring back the idea of a more general capabilities list, then?

The other advantage of a capabilities list is that it is configured fully as part of the roles. Whereas without it, if you want to build something capabilities like, you need to route it through data in the database itself.

text/1029-rbac.rst
```diff
@@ -0,0 +1,152 @@
...
+between reads and writes of objects.
+
+
+Backwards Compatibility
+=======================
+
+Dump/restore of a branch with some existing roles must be implemented such
+that these roles get assigned all capabilities.
+
+Otherwise, this proposal is fully backward compatible.
```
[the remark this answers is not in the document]
aljazerzen: I don't think there is anything special needed for inplace upgrades:

- postgres database does not have to be modified in any way, since we are not creating postgres ROLEs,
- std lib needs to be updated, but only in terms for metadata, which is not written in the database anyway (right?).

aljazerzen:
TODO:
- elvis pointed out that representing capabilities as plain strings is error prone and not introspectable. I see no major downsides as having capabilities be schema objects (similar to futures) that you can define and then check against. Roles would still get capabilities assigned, but that would now have to happen *after* schema of an instance is populated.

aljazerzen:
```
CREATE CAPABILITY Admin;

CREATE CAPABILITY DataExport;

CREATE TYPE Secret {
    CREATE PROPERTY super_secret: str;

    CREATE ACCESS POLICY allow_admin
        ALLOW ALL
        USING (default::Admin in global sys::current_capabilities);

    CREATE ACCESS POLICY allow_warehouse
        ALLOW SELECT
        USING (default::DataExport in global sys::current_capabilities);

    CREATE ACCESS POLICY allow_inserts
        ALLOW INSERT;
};

---

CREATE ROLE aljaz {
  set capabilities := 'Admin,DataExport,DDL,DML';
};

**Comment by elprans** [comment]
```
CREATE PERMISSION Admin;

CREATE PERMISSION DataExport EXTENDING Admin;

CREATE TYPE Secret {
    CREATE PROPERTY super_secret: str;

    CREATE ACCESS POLICY allow_admin
        ALLOW ALL
        USING (sys::have_permission(default::Admin));

    CREATE ACCESS POLICY allow_warehouse
        ALLOW SELECT
        USING (sys::have_permission(default::DataExport));

    CREATE ACCESS POLICY allow_inserts
        ALLOW INSERT;
};

---

CREATE ROLE aljaz {
  SET permissions := {Admin, DataExport, TRANSACTION,SESSION_CONFIG, PERSISTENT_CONFIG};
};
```

Alternatively, we can make permissions as `bool` globals (either via vanilla `global` or a new subvariant of it: `permission`:

```
    CREATE ACCESS POLICY allow_admin
        ALLOW ALL
        USING (permission default::Admin));
```

**Comment by 1st1** [comment]
@elprans so pretty much exact same thing that Aljaz just proposed but with "capabilities" renamed to "permissions", and `permissions` being a set, not a `str`?

What will `TRANSACTION` ident be here? Will it be a hardcoded `CREATE PERMISSION` thing at a system level? What's its fully qualified name? Does it live inside the schema or outside of it?

**Comment by 1st1** [comment]
>Alternatively, we can make permissions as bool globals (either via vanilla global or a new subvariant of it: permission:

Making a subvariant of `global` means we'll have to make `permission` a reserved keyword -- would be nice to avoid that

**Comment by aljazerzen** [comment]
I prefer `sys::has_permission(MyPermission)` over `permission MyPermission`.

---

> What will TRANSACTION ident be here? Will it be a hardcoded CREATE PERMISSION thing at a system level?

Yes. Well, not exactly "hard-coded", it can be part of stdlib:

```
CREATE PERMISSION sys::Transaction;

CREATE ROLE aljaz {
  SET permissions := {
    Admin, DataExport, sys::Transaction, sys::SessionConfig
  };
};

```

**Comment by elprans** [comment]
> I prefer `sys::has_permission(MyPermission)` over `permission MyPermission`.

One problem with this is that it makes `MyPermission` appear as a datum, i.e that you can `select MyPermission`, which is what got me thinking that maybe we should just make them actual data.  But then we probably really want to qualify them explicitly like we do globals (for the same reasons).

**Comment by msullivan** [comment]
What would be the *result* of selecting `MyPermission` (with or without a keyword in front of it)?

Seems kind of odd to me to have that

**Comment by msullivan** [comment]
I also think in general I'm a little unsure about putting it in the schema, since roles are not. What happens when a role references something that gets deleted?

Not putting it in the schema also has big downsides though too

**Comment by msullivan** [comment]
There is a bigger problem with putting permissions in the schema, which is that roles are *instance-wide*, not per-database.

File: text/1029-rbac.rst [hunk 4]
```diff
@@ -0,0 +1,340 @@
...
+To provide access control to functions, we introduce a new function metadata
+field named ``require_permission``. It adds the provided permission to the list
+of required permissions of the query, similar to how `DDL` and other existing
+capabilities work.
+
+code::
+
+    CREATE FUNCTION sys::reset_query_stats(...) -> ... {
+        SET volatility := ...
+        SET require_permission := sys::query_stats;
```
msullivan, line 183: Should this take a set?

(We may need to fiddle with the syntax here from an implementation perspective; right now, `SET` things get evaluated by the static evaluator, which I think will reject this. We'll need to figure out how we want to do it.)

text/1029-rbac.rst
```diff
@@ -0,0 +1,340 @@
...
+defined in ``sys`` module, similar to ``sys::data_modification``.
+
+Before this RFC, non-SUPERUSER roles has practically no permissions. This RFC
+aims to grant as little access as possible, while still being practical.
+That's because granting new access is a non-breaking change, but taking away
+access is a breaking change.
+
+
+Permission inheritance
+----------------------
```
msullivan, line 214: It's possible we might also want to organize this in the opposite direction, with `permission groups`, or something?

msullivan (approved):

text/1029-rbac.rst
aljazerzen: I don't think set would be needed anywhere now, but implementation-wise it is free. So yes, it could be a set, to make it future-proof.

dnwpark:
Are we still adding some kind of `grant_permissions` for things like:
```
type User {
  trigger log_insert after insert for each do (
    insert Log {
      action := 'insert',
      target_name := __new__.name
    }
  );
};
```

where inserting to log is restricted.

dnwpark:
Also, is it maybe better to use `permission sys::data_modification` instead of `global sys::data_modification`?

elprans:
> Are we still adding some kind of `grant_permissions` for things like:
> where inserting to log is restricted.

I think the decision was that we'll add a "security" field which would be an enum of (`Definer`, `Invoker`) a-la SQL and that triggers/functions and other mutation actors can be declared with `security := Definer` to effectively run in superuser context.

dnwpark:
Ok I'll add a little bit in the context role switching section so that we don't forget this.

I'm also going to look at using a `permission` keyword in expressions. `global` is currently a "future reserved keyword" so I think this will probably still work?

msullivan:
> I'm also going to look at using a `permission` keyword in expressions. `global` is currently a "future reserved keyword" so I think this will probably still work?

`permission` won't work, because it will need to be reserved. (Double check, but I think I tried it.)
`global` really should just be an *actual* reserved keyword, not future reserved.

(We ought to reject actually using future reserved keywords.)

dnwpark:
Yeah I tried it and it's exactly as you and elvis said.

In any case, we can always add it as a keyword later if we want.

text/1029-rbac.rst
msullivan: Ah, roles have inheritance, though!

PR #110, opened 2025-04-16
1042: Improved FastAPI workflow

1st1:


anbuzin (approved):

text/1042-fastapi.rst
```diff
@@ -0,0 +1,261 @@
...
+  - ``default.py`` as the last line can have ``from .__default_extras import *``, followed by ``__all__ += (...)``.
+
+* ``app/models/__init__.py`` will re-export all top-level namespaces, so that ``from . import models; print(models.default.User)`` works.
+
+* ``app/models/__init__.py`` will export ``get_db()`` and ``get_async_db()`` factories for sync and async ORM-aware clients respectively.
+
+* Generated Python types will have a special ``__variants__`` class property to access "building bricks" -- fragments of types that can be re-shaped into different useful dataclasses, e.g.:
+
+  - ``models.default.User.__variants__.Empty`` -- an empty type that has ``id`` and ``__tid__`` attributes.
+  - ``models.default.User.__variants__.Optional`` -- a type with all links and properties marked as optiona.
```
fantix, line 79: typo

```suggestion
  - ``models.default.User.__variants__.Optional`` -- a type with all links and properties marked as optional.
```

text/1042-fastapi.rst
```diff
@@ -0,0 +1,261 @@
...
+               class __linkprops__:
+                   weight: int | None
+
+  - This enables link properties to always be fetched and never shadow regular properties or links. Accessing them will be possible via the ``__linkprops__`` property, e.g.:
+
+    .. code-block:: python
+
+        user = await db.get(User.select(friends=User.friends.select()))
+
+        weight = user.friends.__linkprops__.weight  # type: `int | None`
```
fantix, line 172: `user.friends` should be a collection? So

```python
for friend in user.friends:
    weight = friend.__linkprops__.weight  # type: int | None
```

fantix (approved):
This looks good to me overall.

Is my understanding correct here: `db` will be the only interface to actually interact with the database, so there won't be an `await` until you involve the `db` object (assuming async) like `db.query()` or `db.save()`; all `Users.select(...)` or `User(...)` statements will be pure query-building commands that are pending execution. This also means there will be no lazy loading of properties; `db.query()` will simply load everything into memory.

Questions:
* How does the delete command look like? `db.delete(*objects, cascade=True)`?
* How do transactions work? Just a drop-in replacement of the `db` object (but you'll need to commit it with `async with`)?

1st1:
Fix:

* `create_db()` -> `get_db()`

text/1042-fastapi.rst
```diff
@@ -0,0 +1,356 @@
...
+Codegen
+=======
+
+* We're introducing a new tool: ``gel generate``. It will:
+
+  - simplify the implementation of code generators and improve the discoverability of them.
+
+  - be part of our CLI and will be written in Rust. It will delegate the actual code generation to separaate tools. ``gel generate X`` will try the following:
+
+    - check if ``gel-X`` is available in ``PATH``, if so - run it.
```
elprans, line 64:
```suggestion
    - check if ``gel-generate-X`` is available in ``PATH``, if so - run it.
```

text/1042-fastapi.rst
```diff
@@ -0,0 +1,356 @@
...
+get_db() and get_async_db()
+---------------------------
+
+These will be the new API for creating ORM-aware clients. They will also be the *preferred* API.
+
+Users will be importing these from ``app.models``:
+
+.. code-block:: python
+
+    from app.models import create_async_db
```
elprans, line 132:
```suggestion
    from app.models import get_async_db
```

text/1042-fastapi.rst
```diff
@@ -0,0 +1,356 @@
...
+
+    .. code-block:: python
+
+        user = await db.get(User.select(friends=User.friends.select()))
+
+        weight = user.friends.__linkprops__.weight  # type: `int | None`
+
+* ``.id`` is a required property in Gel, however, in Python *new* objects won't have an ``.id`` until they pass through a ``save()`` call. This presents a dilemma: should the ORM type's ``id`` property be optional or required? We say it will be required, because objects will be read more often than written, so we want to optimize for the common case. Accessing the ``id`` property on an unsaved object will raise a runtime error.
+
+  - That said, the reflected types will have a custom ``__init__`` that will not accept an ``id`` argument (in disagreement with the type annotation for the field being "required").
```
elprans, line 200: No disagreement per-se. The non-optional `id` has an implicit default (and we can annotate it as such in Pydantic).

text/1042-fastapi.rst
```diff
@@ -0,0 +1,357 @@
...
+
+  - The custom ``__init__`` will also require passing all required links, which would be in disagreement with all links being "optional" in Python by default.
```
vpetrovykh, line 202: We have server-side defaults for links and props. The links and props themselves might be required, but don't need to be specified because of the server-side default. This should probably use a similar mechanism to the default value for the `id`.

text/1042-fastapi.rst
```diff
@@ -0,0 +1,357 @@
...
+
+* Reflected Gel types will have class methods on them to build queries, e.g. ``User.select(..)``, ``User.filter(..)``, etc.
+
+* Python's ``User.select()`` will be equivalent to ``select User { * }`` in EdgeQL -- this is the only way how we can make Python limited typing work.
+
+* We will also be *always* fetching all link properties by default.
+
+* We'll be adding a new ``lazy`` annotation / field to exclude properties/computeds/link-properties from splats.
+
+* Objects can be mutated and saved: ``user.name = 'Peter'; await db.save(User(name='Anna'), user)``
```
vpetrovykh, line 160: `save` should have atomicity settings, maybe an `atomic=True` parameter:
- `atomic=True`: what it says on the tin, either all objects save successfully or none - no partial data in the database
- `atomic=False`: saving operation will stop at the first error, but any data successfully saved prior to that will make it into the database, no rollbacks

By default saving should prioritise atomicity for consistency reasons. The non-atomic option is more suited for saving objects in bulk where it may be more important to save as much as possible in a large batch because it's preferable to have fewer objects left over to deal with during error recovery. Also, if atomicity is not required, then saving several objects can be done in parallel as long as the connection pool has several connections available (again, useful for bulk data).

text/1042-fastapi.rst
```diff
@@ -0,0 +1,357 @@
...
+        user.email = 'peter@example.com'  # <- fine!
+        await db.save(user)
+
+* "Multi" links and properties will be represented as list-like collections in Python:
+
+  - Properties will be represented as ``gel.List`` (the exact name is still TBD) and will allow duplicate entries.
+
+  - Links will be represented as ``gel.DistinctList`` and will not allow duplicate entries.
+
+  - ``List`` and ``LinkList``, like Python's ``list`` will implement the ``+=`` operator to extend the list, e.g. ``user.friends += [friend]`` (note the square brackets on the right hand side).
```
vpetrovykh, line 239: I assume that here and below `LinkList` is meant to be `DistinctList`. In any case, I would vote strongly against `LinkList` as a name, since it's so similar to a "linked list", which makes it unnecessarily confusing.

text/1042-fastapi.rst
scotttrinh: Yeah, I think in general we have some ambiguity about what "required" means from the user's perspective. In Zod, for instance, we have a concept of "input" vs "output" type that let's defaulted values be optional when parsing input, while reflecting that the value is in fact always there in the parsed output. Not sure if there is some similar concept in Pydantic/Python types.

anbuzin (approved):

text/1042-fastapi.rst
```diff
@@ -0,0 +1,357 @@
...
+
+  - simplify the implementation of code generators and improve the discoverability of them.
+
+  - be part of our CLI and will be written in Rust. It will delegate the actual code generation to separaate tools. ``gel generate X`` will try the following:
+
+    - check if ``gel-X`` is available in ``PATH``, if so - run it.
+
+    - check against the list of known generators, if found - run it. E.g. ``gel generate ts/querybuilder`` can attempt running ``npx @gel/ts-querybuilder``.
+
+  - will provide the necessary CLI connection arguments and will be resolving the instance to to generate code for. Once the instance is resolved, the tool can invoke the actual generator tool as a subprocess, passing the instance name/secrets via env vars.
```
vpetrovykh, line 68:
```suggestion
  - will provide the necessary CLI connection arguments and will be resolving the instance to generate code for. Once the instance is resolved, the tool can invoke the actual generator tool as a subprocess, passing the instance name/secrets via env vars.
```

text/1042-fastapi.rst
```diff
@@ -0,0 +1,357 @@
...
+
+  - will have a ``--list`` flag to list all available known generators.
+
+* The ``.edgeql`` to ``.py`` code gen tools will be updated to to customize the generated code in several ways:
```
vpetrovykh, line 72:
```suggestion
* The ``.edgeql`` to ``.py`` code gen tools will be updated to customize the generated code in several ways:
```

text/1042-fastapi.rst
1st1: Hm, i started updating it and then stumbled upon this (which prompted me to use LinkList in the first place):

>   - ``LinkList`` will have a custom-tailored ``append()`` method that would accept keyword-only arguments to set link properties, e.g. ``user.friends.append(friend, weight=10)``.

So `DistinctList` wouldn't work. It should be either `LinkList` or we coin a better name. The silver lining here is that the name of the class is kinda internal here, users shouldn't care or see it ever.

PR #113, opened 2025-05-09
Update to how `gel watch` should interact with other tools.

vpetrovykh:


mmastrac (approved):
I believe we have implemented most of this change.

PR #120, opened 2025-06-24
Support restricting a role to certain branches

msullivan:


text/1029-rbac.rst
```diff
@@ -240,6 +242,23 @@ Requires ``cfg::perm::configure_allow_user_specified_id``
 The configuration will be done with annotations on Config objects,
 like everything else with the config system.
 
+Data branch restriction
+-----------------------
+
+Roles are also extended with a ``branches`` field, enumerating the branches
+that the role is allowed to connect to. The default value is ``{ '*' }``,
+signifying all branches.
```
aljazerzen, line 250: We don't allow creating a branch named `*`, right?

dnwpark (approved):

text/1029-rbac.rst
```diff
@@ -240,6 +242,23 @@ Requires ``cfg::perm::configure_allow_user_specified_id``
...
+
+With the right permissions granted (
+``sys::perm::data_modifiction``, ``sys::perm::ddl``,
+``sys::perm::branch_config``), this allows a user to be granted substantial
+control over a single branch without being able to make changes elsewhere.
+
+(It is actually basically already possible to accomplish this, but
+only when using JWT based authentication with scopes.)
+
+The branches field is *not* inherited in any way.
```
dnwpark, line 260: Maybe we ought to make roles with branches non-inheritable in some way?

text/1029-rbac.rst
aljazerzen: You can use ` to escape any char sequence to be an identifier. There are a few exceptions, but I don't think `*` is one of them:

```
_localdev:main> create empty branch `*`;
OK: CREATE BRANCH
```

So we have a few options:
- this is allowed, if a user has a branch named `*` and adds `*` to a role's branches, the role has access to all branches. So they cannot restrict access to just `*`. It's a weird thing to do anyway.
- we forbid creating branches named `*`,
- we forbid all identifiers from containing `*`,

I would actually prefer first option.

